PR #1, opened 2023-04-09
Get SQLite manifest

gwaybio:
Search through JUMP data on Cell Painting gallery (https://cellpainting-gallery.s3.amazonaws.com/index.html#cpg0016-jump/) to obtain full paths to SQLite files


0.convert/nbconverted/0.sqlite_manifest.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Use client to list source directories
+# Sources represent different sites from the JUMP consortium
+client = boto3.client("s3")
+
+top_level_jump_result = client.list_objects(
+    Bucket=s3_bucket_name,
+    Prefix=f"{jump_id}/",
+    Delimiter="/"
+)
```
d33bs, line 56: Minor non-required suggestion: Would it make sense to take advantage of [`cloudpathlib.s3path()`](https://cloudpathlib.drivendata.org/stable/api-reference/s3path/) to keep the syntax in alignment with `pathlib` throughout?

0.convert/nbconverted/0.sqlite_manifest.py
```diff
@@ -0,0 +1,145 @@
...
+perturbation_type_url = f"{jump_github}/raw/{jump_data_commit}/metadata/plate.csv.gz"
+
+output_file = pathlib.Path("metadata", "jump_sqlite_aws_file_locations.tsv")
+
+
+# In[3]:
+
+
+# Connect to Cell Painting gallery S3 session
+# Note: you must have already configured awscli with credentials
```
d33bs, line 38: Are credentials required to view or work with the data? If so, consider providing some mention about what credentials are used (in the broad sense) and how one might obtain permission to the data. If not, consider mentioning that this is publicly available data and possibly extend the code to use anonymous/unsigned mechanisms for reproducibility (and avoid awscli configurations).

0.convert/nbconverted/0.sqlite_manifest.py
```diff
@@ -0,0 +1,145 @@
...
+            # Build SQLite file
+            sqlite_file = f"s3://{s3_bucket_name}/{prefix_string_per_batch}{plate}/{plate}.sqlite"
+            source_plate_info.append([source, batch, plate, sqlite_file])
+
+
+# In[7]:
+
+
+# Compile JUMP data manifest
+jump_df = pd.DataFrame(source_plate_info, columns=["soure", "batch", "plate", "sqlite_file"])
```
d33bs: Double checking: was "source" intended here as a column name?
```suggestion
jump_df = pd.DataFrame(source_plate_info, columns=["source", "batch", "plate", "sqlite_file"])
```

0.convert/nbconverted/0.sqlite_manifest.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Get plate ids
+source_plate_dictionary = {}
+source_plate_info = []
+for source in source_batch_dictionary:
+    source_plate_dictionary[source] = {}
+    source_batches = source_batch_dictionary[source]
+    
+    for batch in source_batches:        
+        prefix_string_per_batch = f"{jump_id}/{source}/{backend_directory}{batch}/"
```
d33bs, line 95: Should there be an additional slash here (acknowledging the name of the `backend_directory` as only a directory and not a prefix)?
```suggestion
        prefix_string_per_batch = f"{jump_id}/{source}/{backend_directory}/{batch}/"
```

environment.yml
```diff
@@ -0,0 +1,19 @@
+name: jump_sc
+channels:
+  - conda-forge
+dependencies:
```
d33bs, line 4: Consider adding `boto3` as an explicit dependency of the project (I think this may be available from `cytotable`). Doing this might not make much of a difference but could help with understanding of the environment.

0.convert/nbconverted/0.sqlite_manifest.py
```diff
@@ -0,0 +1,145 @@
...
+source_batch_dictionary
+
+
+# In[6]:
+
+
+# Get plate ids
+source_plate_dictionary = {}
+source_plate_info = []
+for source in source_batch_dictionary:
```
d33bs: Consider the use of `source_batch_dictionary.items()` here to explicitly handle the key and value within the loop.

0.convert/nbconverted/0.sqlite_manifest.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Append perturbation type to this dataframe
+# Load metadata from https://github.com/jump-cellpainting/datasets
+perturbation_type_df = pd.read_csv(perturbation_type_url)
+perturbation_cols = ["Metadata_Source", "Metadata_Batch", "Metadata_Plate"]
+
+jump_df = (
+    jump_df.merge(
+        perturbation_type_df,
+        left_on=["soure", "batch", "plate"],
```
d33bs: Double checking: should this be "source" instead?
```suggestion
        left_on=["source", "batch", "plate"],
```

d33bs (approved):
Nice work! Overall I thought this looked good and only had suggestions for potential improvements. Please don't hesitate to let me know if you have any questions.

0.convert/nbconverted/0.sqlite_manifest.py
gwaybio: hmm, I don't think i'm following. I tried s3path on the Prefix argument, but it failed with a `ParamValidationError`.

I'm sure I'm getting something wrong here, but i do like the idea of keeping with `pathlib` syntax.

```python
ParamValidationError: Parameter validation failed:
Invalid type for parameter Prefix, value: s3:///cpg0016-jump/, type: <class 'cloudpathlib.s3.s3path.S3Path'>, valid types: <class 'str'>
```

0.convert/nbconverted/0.sqlite_manifest.py
gwaybio: yep! nice catch

0.convert/nbconverted/0.sqlite_manifest.py
gwaybio: `backend_directory = "workspace/backend/"`, which is required for the boto3 call. I probably could dig deeper to make this backslash necessary, but to save time I instead added a comment for clarity

environment.yml
gwaybio: yep, good idea!

0.convert/nbconverted/0.sqlite_manifest.py
gwaybio: yep

0.convert/nbconverted/0.sqlite_manifest.py
gwaybio: nice, thanks for this suggestion

0.convert/nbconverted/0.sqlite_manifest.py
gwaybio: I added an issue #2 - I agree this is a great idea, but I don't think it is necessarily required at the moment. I also don't have the exact knowledge to implement this on the tip of my fingers, so it will take some time to investigate, time that I think is more effectively spent elsewhere, at the moment.

Thanks for raising!

gwaybio:
going to merge! thanks again @d33bs

PR #3, opened 2023-05-17
Convert s3 sqlite to parquet with cytotable

gwaybio:
See cytomining/CytoTable#43 cytomining/CytoTable#52


gwaybio:
I'm merging this despite it not being in final form, in order to make sprint progress

PR #8, opened 2023-08-21
Compare treatment phenotypic expression

MattsonCam:
In this pr, I rename the Dunn statistical test and create volcano plots to represent phenotypic probability significance for each treatment as compared to DMSO. The data for these plots was also included. Please let me know if you have any suggestions or questions.


3.analyze_data/figures/compound/Apoptosis/final/dunn/dunn_volcano_Apoptosis_compound.png
jenna-tomkinson, line 1: This is looking pretty promising that there is a big shift from 0 in the positive direction as I assume that means there a big shift from DMSO for many compound which I hope to see are connected to cell death 🤞

3.analyze_data/nbconverted/make_volcano_predicted_probabilities.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# # Compare phenotypic expression among treatments with volcano plot visualizations
+# We compare phenotypic expression for each (phenotype, statistical test, model type, treatment type) combination
+
+# In[ ]:
```
jenna-tomkinson: Make sure to have run the notebook for this analysis prior to converting to python file so that the code cells are ran in order.

3.analyze_data/nbconverted/make_volcano_predicted_probabilities.py
```diff
@@ -0,0 +1,133 @@
...
+
+# Remove the zero values after the p-value adjustment
+compdf = compdf.loc[compdf["p_value"] > 0]
+significance_threshold = -np.log10(alpha)
+
+# Define the unique groups to plot
+unique_comparisons = set(zip(compdf["phenotype"], compdf["treatment_type"], compdf["model_type"], compdf["statistical_test"]))
+
+# Specify colors according to categories
+colors = {
```
jenna-tomkinson: I think it might be good to include a legend for these plots as to what these colors mean. To someone just looking at this plot without context is a bit hard to follow and the title is pretty long also.

Maybe it would help to have a legend that shows Red = Significant and differential from DMSO.

3.analyze_data/nbconverted/make_volcano_predicted_probabilities.py
```diff
@@ -0,0 +1,133 @@
...
+    'y_below_threshold',
+    axis=1
+)
+
+for comp in unique_comparisons:
+
+    fig, ax = plt.subplots(figsize=(15, 10))
+    groupdf = compdf.copy()
+
+    # Seperate the croups we are intersted in plotting
```
jenna-tomkinson:
```suggestion
    # Separate the groups we are interested in plotting
```

3.analyze_data/nbconverted/make_volcano_predicted_probabilities.py
```diff
@@ -0,0 +1,133 @@
...
+        x="comparison_metric_value",
+        y="neg_log_p_value",
+        hue="category",
+        palette=colors,
+        legend=False,
+        ax = ax
+    )
+
+    # Add threshold lines
+    plt.axhline(
```
jenna-tomkinson: It might also be good to include the vertical threshold lines as well like traditional volcano plots (see example):

image

It would also help to really see what compounds are most significantly different from DMSO. 

If it isn't possible in this application, no worries, but it would really drive home the point of significant compounds that we should look at.

README.md
```diff
@@ -34,4 +34,4 @@ This data in addition to the other single-cell metadata is stored as a parquet f
 To understand each treatments ability to induce each of the 15 phenotypes, we perform post hoc testing with the probabilities between each of the treatments and the control (DMSO).
 These probabilities were generated in `2.evaluate_data`.
 In each of these tests we compute the median to determine the directionality of significance among treatments.
```
jenna-tomkinson, line 36: Should you explain why you use both Dunn and Mann Whitney tests?

Or should this be included in a README for the `3.analyze_data` module?

README.md
```diff
@@ -34,4 +34,4 @@ This data in addition to the other single-cell metadata is stored as a parquet f
...
-We later use the Bonferonni correction to adjust the significance level in the visualization step of the anlaysis.
```
jenna-tomkinson, line 37: Is there a reason you removed this? Is this not used as I did see your use Bernoulli adjustment in the file so would you reference that instead?

jenna-tomkinson: Also, you might want to consider moving this into the main README for the module instead if this comment does still apply.

environment.yml
```diff
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - conda-forge::python=3.9
-  - pandas
+  - pandas<2.0
```
jenna-tomkinson, line 6: I would recommend having the `conda-forge::` notation for all dependencies in the environment to stay consistent as well as this is the standard it seems for all projects now.

environment.yml
```diff
@@ -13,6 +13,8 @@ dependencies:
   - requests
   - seaborn
   - scikit-posthocs
+  - notebook
+  - ipython
   - pip
   - pip:
     # For creating my python environment:
```
jenna-tomkinson, line 20: Recommend removing this comment if it isn't necessary.

jenna-tomkinson (changes requested):
Great PR and I love all of the figures! 😄 

I do have a few comments I would like addressed:

The main thing I am curious about is why you choose to use both Dunn and Mann Whitney to perform post-hoc analysis. Based on visualization alone, figures for both of these tests (in both shuffled and final models) look the exact same. Like it seems like every point is in the exact same place. This could also be a fluke and something happened in the figure generation process. If not, I was wondering why we would need both analysis if the results are the same?

As well, I mention about adding vertical threshold lines (if possible). 

I would recommend adding a README for this module so that you can explain things like why you perform two different post-hoc tests and how to run the notebooks, etc.

PR #12, opened 2023-10-18
Add conda_forge to all dependencies

jenna-tomkinson:
# Add conda_forge to all dependencies

In this small PR, I add the `conda_forge` prefix to all dependencies in the environment file to match the standard Way Lab conventions for conda env files.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,18 +3,18 @@ channels:
   - conda-forge
 dependencies:
   - conda-forge::python=3.9
-  - pandas<2.0
-  - numpy
-  - matplotlib
-  - scikit-learn=1.1.1 # Version to allow Roshan's sklearn models to predict probabilities correctly
-  - scipy=1.8.1
-  - pathlib2
-  - scikit-learn
-  - requests
-  - seaborn
-  - scikit-posthocs
-  - notebook
-  - ipython
-  - pip
+  - conda-forge::pandas<2.0
+  - conda-forge::numpy
+  - conda-forge::matplotlib
+  - conda-forge::scikit-learn=1.1.1 # Version to allow Roshan's sklearn models to predict probabilities correctly
+  - conda-forge::scipy=1.8.1
+  - conda-forge::pathlib2
+  - conda-forge::scikit-learn
+  - conda-forge::requests
+  - conda-forge::seaborn
+  - conda-forge::scikit-posthocs
+  - conda-forge::notebook
+  - conda-forge::ipython
+  - conda-forge::pip
   - pip:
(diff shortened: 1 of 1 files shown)


environment.yml
```diff
@@ -3,18 +3,18 @@ channels:
...
-  - scipy=1.8.1
-  - pathlib2
-  - scikit-learn
-  - requests
-  - seaborn
-  - scikit-posthocs
-  - notebook
-  - ipython
-  - pip
+  - conda-forge::pandas<2.0
```
MikeLippincott: Consider adding a comment for why <2.0 ( I know why, but might someone looking at this not know why?)

environment.yml
```diff
@@ -3,18 +3,18 @@ channels:
...
+  - conda-forge::numpy
```
MikeLippincott, line 7: Similar to pandas limiting the version might be helpful here. There are big differences between versions of numpy specifically with data type specifications.

environment.yml
```diff
@@ -3,18 +3,18 @@ channels:
...
+  - conda-forge::matplotlib
+  - conda-forge::scikit-learn=1.1.1 # Version to allow Roshan's sklearn models to predict probabilities correctly
```
MikeLippincott, line 9: Good comment here explains the version

MikeLippincott (approved):
LGTM, feel free to address or ignore comments. The versions might be helpful to prevent analysis from becoming non-reproducible upon library updates

environment.yml
jenna-tomkinson: @MikeLippincott I am not sure why, can you explain? 

I know pandas 2.0 has been causing issues with other dependencies like sqlalchemy but I am not sure about any other issues.

environment.yml
jenna-tomkinson: @MikeLippincott 

Does it make sense to limit it to the version Cameron's environment uses (e.g., his version of the env he installed uses numpy v1.24.4)?

environment.yml
MikeLippincott: Exactly  why. It causes issues. The semantic versioning had a major update so some aspects of the API are different.

environment.yml
MikeLippincott: yeah that could work, but generally on the environment files we prefer to keep the versions slightly flexible. So instead of one version (unless it only works on one version) try using a `<` or `>` or `=<` etc

jenna-tomkinson: After discussing with @MattsonCam, he believes it would be best not to include this as to avoid any other conflicts in the future.

environment.yml
jenna-tomkinson: Added code comment! 👍

jenna-tomkinson:
@MikeLippincott Thanks for the review! I will be merging now!

PR #19, opened 2024-02-08
Add UMAP analysis for JUMP + Mitocheck

jenna-tomkinson:
# Add UMAP analysis for JUMP + Mitocheck

In this PR, I include refactored code from my local machine where I created UMAPs for different splits/combos of the JUMP or JUMP+Mitocheck data. For each data split (e.g., only JUMP data or JUMP+Mitocheck for only AreaShape features, etc.), I apply the probabilities from either Greg's AreaShape ML model or an original model using all nuclei features from Cameron.

**BIG NOTE:** The pathing for this will be a mess since I got all of the data from DropBox. Also, any intermediate files put into a `data` folder were not uploaded to GitHub due to their very large size (>5 GB).

The main things that need to be reviewed are the code and figures. The large number of lines of code is due to uploading the intermediate UMAP embedding files.


.gitignore
```diff
@@ -12,3 +12,7 @@ __pycache__/
 
 # Ignore mac related files
 *.DS_Store
+
+# Ignore data folder in UMAP analysis
```
gwaybio:
```suggestion
# Ignore data folder in UMAP analysis (>5GB)
```

Good to provide a reason for why. Is my suggestion accurate?

3.analyze_data/UMAP_analysis/nbconverted/1.fs_all_JUMP_plates.py
```diff
@@ -0,0 +1,162 @@
...
+        num_rows = parquet_file.num_row_groups
+
+        # Randomly select a row group from the file
+        random_row_group = random.randint(0, num_rows - 1)
+
+        # Read the selected row group with the specified number of rows
+        df = parquet_file.read_row_group(random_row_group, columns=parquet_file.schema.names, use_threads=True).to_pandas()
+
+        # Randomly sample rows from the selected row group
+        sampled_rows = df.sample(n=1000, random_state=0)
```
gwaybio: Probably ok to keep as is (no change required), but I usually recommend to set these variables (`n_sample`) in the beginning of notebooks

3.analyze_data/UMAP_analysis/nbconverted/1.fs_all_JUMP_plates.py
```diff
@@ -0,0 +1,162 @@
...
+# Remove features not seen in JUMP from Mitocheck feature list
+feature_cols = [item for item in feature_cols if item not in diff_columns]
+
+# Add metadata with nuclei features from Mitocheck
+filtered_df = concatenated_df[metadata_cols + feature_cols]
+
+# Save the concatenated DataFrame as a Parquet file
+filtered_df.to_parquet(filter_save_path, index=False)
+
+print(filtered_df.shape)
```
gwaybio, line 163: This is great! Can I interpret this to be the "number of features in common between JUMP and mitocheck"?

3.analyze_data/UMAP_analysis/nbconverted/2.jump_UMAP.py
```diff
@@ -0,0 +1,389 @@
...
+mito_cp_df.head()
+
+
+# ### Filter the mitocheck data
+
+# In[9]:
+
+
+# Load in labeled mitocheck data
+label_df = pd.read_csv(label_data_path)
```
gwaybio: Looks like you're reading this file in twice?

gwaybio: (line 187)

3.analyze_data/UMAP_analysis/nbconverted/2.jump_UMAP.py
```diff
@@ -0,0 +1,389 @@
...
+
+# load in mitocheck filtered data
+filter_df = pd.read_parquet(pathlib.Path("./data/concat_mitocheck_data_subset.parquet"))
+
+# load in predicted probabilites file
+prob_path = pathlib.Path("./data/all_features_probabilities.parquet")
+prob_df = pd.read_parquet(prob_path)
+
+# Mitocheck labelled dataset to use to find nuclei features and concat
+label_data_path = "https://github.com/WayScience/phenotypic_profiling_model/raw/main/0.download_data/data/labeled_data__ic.csv.gz"
```
gwaybio: When you're reading from a github url, it's generally advised to load a _versioned_ url. You can find the versioned url by pressing `y` when you're looking at any GitHub file

3.analyze_data/UMAP_analysis/nbconverted/2.jump_UMAP.py
```diff
@@ -0,0 +1,389 @@
...
+mito_cp_df.head(2)
+
+
+# ## Concat mitocheck and jump data (all nuclei features)
+
+# In[10]:
+
+
+# Concatenate the two DataFrames vertically
+merged_mito_jump_df = pd.concat([mito_cp_df, jump_df], axis=0)
```
gwaybio, line 262: Is there a way to confirm that the two datasets have the same column order? Will concat respect order?

One approach to confirm this is to find the overlapping columns and then call `.reindex(columns=common_columns)` (or something like this) to confirm column order.

3.analyze_data/UMAP_analysis/nbconverted/2.jump_UMAP.py
```diff
@@ -0,0 +1,389 @@
...
+    model_dir.mkdir(exist_ok=True)
+
+    # Generate output file and save
+    output_umap_file = pathlib.Path(
+        f"./{model_dir}/{data_name}_{model_type}_{'_'.join(prob_path.stem.split('_')[:2])}_model.tsv"
+    )
+    umap_with_metadata_df.to_csv(output_umap_file, index=False, sep="\t")
+
+print(umap_with_metadata_df.shape)
+umap_with_metadata_df.head()
```
gwaybio, line 386: Great! So just to confirm, all UMAPs you describe here will have output coordinates with Metadata so that I can create figures in the phenotypic_profiling repo?

3.analyze_data/UMAP_analysis/nbconverted/2.jump_UMAP.py
```diff
@@ -0,0 +1,389 @@
...
+
+
+# In[ ]:
```
gwaybio: extra cell to be deleted

3.analyze_data/UMAP_analysis/nbconverted/3.vis_mito_jump.r
```diff
@@ -0,0 +1,202 @@
+suppressPackageStartupMessages(library(dplyr))
```
gwaybio, line 1: you have become an R expert!!! 🥇

gwaybio (approved):
This looks great @jenna-tomkinson ! I made only minor comments; one should definitely be checked (about column alignment in a pd.concat() call), and the others are otherwise super minor. Kudos for a great PR!

.gitignore
jenna-tomkinson: 100% accurate, I have included this now!

3.analyze_data/UMAP_analysis/nbconverted/1.fs_all_JUMP_plates.py
jenna-tomkinson: I updated this! Super small and I get why it makes sense.

3.analyze_data/UMAP_analysis/nbconverted/1.fs_all_JUMP_plates.py
jenna-tomkinson: Correct! We are filtering out any nuclei features in the JUMP that aren't seen in the Mitocheck data so we only keep similar morphology features.

3.analyze_data/UMAP_analysis/nbconverted/2.jump_UMAP.py
jenna-tomkinson: Weird, great catch! I guess I do. I will remove the later once since it was loaded in earlier. Thanks!

3.analyze_data/UMAP_analysis/nbconverted/2.jump_UMAP.py
jenna-tomkinson: Just added this! Good thing to know for the future, I never knew about this at all!

3.analyze_data/UMAP_analysis/nbconverted/2.jump_UMAP.py
jenna-tomkinson: Correct! All of these will have UMAP coordinates and associated in Metadata expect in the Mito+JUMP data split, the only Metadata is the predicted class, predicted value, and data set since the JUMP Metadata didn't match with Mito.

3.analyze_data/UMAP_analysis/nbconverted/2.jump_UMAP.py
jenna-tomkinson: Good catch! Removed.

3.analyze_data/UMAP_analysis/nbconverted/3.vis_mito_jump.r
jenna-tomkinson: Thank you! 🎉 I will only partially accept this award until I figure out patchwork and THEN I will truly feel like an expert 😆

3.analyze_data/UMAP_analysis/nbconverted/2.jump_UMAP.py
jenna-tomkinson: This definitely would respect column order to the first data frame. I know that CellProfiler outputs will always put the columns in alphabetical order (based on feature) and I can see that the order of the JUMP data and Mito data were in the same order.

I also double checked with this line `mito_cp_df.columns.tolist() == jump_df.columns.tolist()` and they are the same order (True).

jenna-tomkinson:
@gwaybio Thank you lots for the review! I am merging now! Let me know if you have any other questions regarding the code and I am excited to see more figures for the paper!

PR #23, opened 2024-03-06
Aggregate phenotype probabilities across well

MattsonCam:
In this pr, the phenotype probabilities both the shuffled and final models are aggregated by well. I may decide to refactor this code to improve runtime. Otherwise please let me know if you have any suggestions.


3.analyze_data/nbconverted/log_reg_class_balanced_areashape_aggregate_well_predicted_probabilities.py
```diff
@@ -0,0 +1,259 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# # Compare Well Treatements
+# We compare the treatments in each well using cell treatment probabilities and negative control probabilities for each phenotype.
+# This comparison is accomplished with a KS Test.
```
gwaybio: It looks like this documentation is incorrect? Maybe a copy+paste issue from the KS test notebook?

3.analyze_data/nbconverted/log_reg_class_balanced_areashape_aggregate_well_predicted_probabilities.py
```diff
@@ -0,0 +1,259 @@
...
+
+import pathlib
+import sys
+
+import numpy as np
+import pandas as pd
+
+# Import significance test utils
+sys.path.append("utils")
+import well_significance_testing as sig_test
```
gwaybio, line 18: potentially another artifact of KS testing

3.analyze_data/nbconverted/log_reg_class_balanced_areashape_aggregate_well_predicted_probabilities.py
```diff
@@ -0,0 +1,259 @@
...
+
+# In[4]:
+
+
+comparison_results_output_filename = "class_balanced_well_log_reg_areashape_aggregate_probabilities.parquet"
+output_path = pathlib.Path("class_balanced_well_log_reg_aggregate_probabilities")
+output_path.mkdir(parents=True, exist_ok=True)
+
+
+# ## KS test wrapper function
```
gwaybio: Another artifact possibly

3.analyze_data/nbconverted/log_reg_class_balanced_areashape_aggregate_well_predicted_probabilities.py
```diff
@@ -0,0 +1,259 @@
...
+tracked_cols = ["treatment", "treatment_type", "Time"]
+
+# Store phenotype column names
+phenotype_cols = None
+
+# Store comparison data
+comparisons = None
+
+# Compute comparisons by iterating through each plate
+for proba_path in list(proba_plate_paths):
```
gwaybio, line 222: you might consider hiding runtime warnings - however, it is totally fine (probably even encouraged!) to skip any modifications that require rerunning this notebook. Any optimization, or minor upgrades (for example, hiding runtime warnings), is not necessarily worth pursuing given the output will not change.

3.analyze_data/nbconverted/log_reg_class_balanced_areashape_aggregate_well_predicted_probabilities.py
```diff
@@ -0,0 +1,259 @@
...
+# ## Compare treatments and negative controls
+
+# In[10]:
+
+
+# Define columns to group by
+filt_cols = ['Metadata_Plate', 'Metadata_Well', 'Metadata_model_type', 'Cell_type']
+
+# Columns of interest which should also be tracked
+tracked_cols = ["treatment", "treatment_type", "Time"]
```
gwaybio: I don't see this variable used - also, I downloaded the aggregated probabilities and I don't see these columns (I do end up seeing cell type!)

3.analyze_data/nbconverted/log_reg_class_balanced_areashape_aggregate_well_predicted_probabilities.py
gwaybio: Given the comment above about not seeing the important variables, it seems you will have to rerun this notebook. However, I would recommend making only the necessary fixes to obtain the correct data with the lowest amount of time/pain spent optimizing since this notebook will not need to be run over and over again (presumably)

gwaybio (approved):
Looks great @MattsonCam ! I made a couple documentation suggestions.

The main comment is that the output still does not have the necessary metadata columns. I made specific in-line comments where I suspect the issue is happening.

I also want to bring to your attention that 302 treatments have NA values for all phenotype probabilities. These 302 treatments do not have an apparent pattern (not all from the same well or plate) - so please double check this too! (we did not have this issue with the last aggregated probabilities file)

I'll approve this PR now so you can feel free to merge in once you've made the necessary changes. Thanks!

MattsonCam:
Hey @gwaybio, it looks like the tracked columns should be included now. If there is another column I should include, just let me know. Otherwise, I will go ahead and merge this pr.

gwaybio:
> If there is another column I should include, just let me know. Otherwise, I will go ahead and merge this pr.

There are no other columns I can think to add now. Please merge at will!

MattsonCam:
Thanks @gwaybio, merging now!

PR #26, opened 2024-03-20
Allowing to download JUMP dataset without credentials 

axiomcura:
This PR updates the AWS download bash script by removing the requirement for providing credentials.


gwaybio (approved):

MattsonCam (approved):
Thanks for adding this @axiomcura, merging now!

PR #29, opened 2024-09-30
Add `justfile` for env management and common project commands

d33bs:
This PR adds a `justfile` which may be used through [`just`](https://github.com/casey/just) for performing common project commands. I'm adding this to help account for changes over time and to save time when it comes to documentation + reproducibility of the project. I expect to expand this file with other commands but wanted to start with existing needs first and get early input on the format / usage.

Thanks for any thoughts or feedback you may have!

### DIFF
--- a/justfile
+++ b/justfile
@@ -3,19 +3,32 @@
 
 # find system default shell
 hashbang := if os() == 'macos' {
-	'#!/usr/bin/env zsh'
+	'/usr/bin/env zsh'
 } else {
-	'#!/usr/bin/env bash'
+	'/usr/bin/env bash'
 }
 
 # show a list of just commands for this project
 default:
   @just --list
 
-# setup conda envs
+# setup conda envs for this project
 @setup-conda-envs:
     #!{{hashbang}}
-    conda remove -n jump_sc --all
-    conda remove -n R_jump_sc --all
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -0,0 +1,21 @@
...
+
+# show a list of just commands for this project
+default:
+  @just --list
+
+# setup conda envs
+@setup-conda-envs:
+    #!{{hashbang}}
+    conda remove -n jump_sc --all
+    conda remove -n R_jump_sc --all
```
jenna-tomkinson: Why have these remove sections? Why not just only have the install environments?

jenna-tomkinson (approved):
LGTM! I added one question for you to address but other than that this is a super simple and nice addition!

justfile
d33bs: Thanks for this - it's a good question! I've encountered behavior from conda in the past that made it so any changes to the `environment.yml` could potentially corrupt or have erroneous behavior when modifying the existing environments. The removals here were a matter of trying to be safe about this. That said, I might be being overprotective here, and will shift to use something like `conda env update --name my_env --file environment.yml --prune` instead.

d33bs:
Thanks @jenna-tomkinson for your review! I've made a small adjustment to help account for your comment. Merging this in now.

Issue #30, opened 2024-10-01
Process JUMP pilot data (`cpg0000-jump-pilot`) with CytoTable


d33bs:

d33bs:
Plates listed [here](https://github.com/WayScience/JUMP-single-cell/blob/main/0.download_data/data/jump_dataset_location_manifest.csv) have now been processed within a local lab computing environment. Processing took roughly 7 days.

PR #33, opened 2024-10-03
Add CytoTable processing of JUMP plate data

d33bs:
This PR adds a Python script which can be used to process JUMP plate data using CytoTable. Along the path towards this PR I also made adjustments to existing scripts and files. Notably, I made a change which assumes that all steps can be run from the root of the repo using `just`, which I feel will simplify how scripts are written. I've added a subshell example within the `run.sh` script for `0.download_data` for steps which must retain their relative pathing.

Please note: I've added a `pre-commit-config.yaml` file (and related `pyproject.toml` file for `ruff` configuration) to the project which included many linting tools that I found helpful during development. As a result of this the `README.md` has been updated in sections that don't apply to the intended goal of this PR.

I haven't processed the modifications found within this step, this PR is intended to provide a preview of intent which I'll run afterwards.

Thanks in advance for any thoughts or feedback on this work!

References #30


0.download_data/1.process_JUMP_plates_with_CytoTable.py
```diff
@@ -0,0 +1,71 @@
...
+from pyarrow import parquet
+
+# create plates directory if it doesn't exist already
+if not (plates_folder := pathlib.Path("./0.download_data/data/plates")).is_dir():
+    plates_folder.mkdir()
+
+# process each plate from the manifest individually
+for _, plate_name, plate_s3_path in pd.read_csv(
+    "./0.download_data/data/jump_dataset_location_manifest.csv", header=0
+).to_records():
```
MattsonCam, line 23: I like the way you did this

0.download_data/1.process_JUMP_plates_with_CytoTable.py
```diff
@@ -0,0 +1,71 @@
...
+    print("Processing plate ", plate_name)
+
+    # create a folder for the plate
+    if not (
+        plate_folder := pathlib.Path(f"./0.download_data/data/plates/{plate_name}")
```
MattsonCam, line 28: walrus :clap:

MattsonCam: Maybe you really want to make sure the all of the directories already exist, but if not, you could consider using:
`plate_folder = pathlib.Path(f"./0.download_data/data/plates/{plate_name}")
plate_folder.mkdir(parents=True, exist_ok=True)`

to create the parent directories. This also doesn't seem to override the directories if they already exist

0.download_data/1.process_JUMP_plates_with_CytoTable.py
```diff
@@ -0,0 +1,71 @@
...
+            f"./0.download_data/data/plates/{plate_name}/{plate_name}.parquet"
+        )
+    ).is_file():
+        # process plate using CytoTable
+        cytotable_output_path = cytotable.convert(
+            source_path=plate_s3_path,
+            dest_path=cytotable_output_path,
+            dest_datatype="parquet",
+            source_datatype="sqlite",
+            chunk_size=8000,
```
MattsonCam, line 44: Is this a practical parameter?

0.download_data/1.process_JUMP_plates_with_CytoTable.py
```diff
@@ -0,0 +1,71 @@
...
+            preset="cellprofiler_sqlite_cpg0016_jump",
+            # allows AWS S3 requests without login
```
MattsonCam, line 46: Cool to see this implemented

0.download_data/1.process_JUMP_plates_with_CytoTable.py
```diff
@@ -0,0 +1,71 @@
...
+            no_sign_request=True,
```
MattsonCam, line 47: I'm not seeing this parameter in the function signatures of cytotable v.0.0.10, unless it's not used until three or more method calls with kwargs? Something maybe worth checking

0.download_data/1.process_JUMP_plates_with_CytoTable.py
```diff
@@ -0,0 +1,71 @@
...
+            # use explicit cache to avoid temp cache removal
+            local_cache_dir="./0.download_data/jump_sqlite_s3_cache/",
```
MattsonCam, line 49: Same comment here. Is the variable specified in a CloudPath class or further down the line?

0.download_data/run.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# convert all notebooks to python files into the nbconverted folder
+conda run -n jump_sc jupyter nbconvert --to python \
+    --output-dir=./0.download_data/nbconverted/ \
+    ./0.download_data/*.ipynb
+
+# download the jump manifest data
+# use a subshell to keep existing code intact
+(
+cd ./0.download_data || exit
```
MattsonCam, line 13: Could also use a trap for more flexibility in error handling

justfile
MattsonCam, line 1: I've seen a similar rust tool used in a linux setup project. What is the advantage of this over a make file, for example?

MattsonCam (approved):
Just left a few comments. Wasn't sure if some of the cytotable.convert function parameters are defined, but you would know better than I would. It may help to perform a test run if you haven't performed one already. Otherwise, LGTM, good job!

0.download_data/1.process_JUMP_plates_with_CytoTable.py
d33bs: Cheers thanks!

0.download_data/1.process_JUMP_plates_with_CytoTable.py
d33bs: Thanks! I'll add in the `parents` and `exist_ok` parameters, good idea.

0.download_data/1.process_JUMP_plates_with_CytoTable.py
d33bs: Thanks for checking on this - yes, I feel this is practical for the time being. I used similar within earlier work with the JUMP data which may be found here: https://github.com/WayScience/coSMicQC/blob/main/docs/src/examples/jump_umap_analysis_with_cosmicqc.py . I wonder if it would make sense to test this more generally to see how performance is impacted but hesitate to change something that worked in earlier iterations.

0.download_data/1.process_JUMP_plates_with_CytoTable.py
d33bs: Thanks, yes, agreed!

0.download_data/1.process_JUMP_plates_with_CytoTable.py
d33bs: Thanks for checking on this. This is passed through `**kwargs` to reach [`cloudpathlib.S3Client`](https://cloudpathlib.drivendata.org/stable/api-reference/s3client/). We document this behavior within the [`cytotable.convert` docstring](https://github.com/cytomining/CytoTable/blob/main/cytotable/convert.py#L1362) and also within [the docsite](https://cytomining.github.io/CytoTable/overview.html#cloud-service-configuration-and-authentication).

0.download_data/1.process_JUMP_plates_with_CytoTable.py
d33bs: Thanks for checking on this again here. Please see [reply here](https://github.com/WayScience/JUMP-single-cell/pull/33#discussion_r1792207975).

justfile
d33bs: Thanks for asking about this - I think a `Makefile` could have been used here but generally I recommend and prefer to use tools which were purpose-built for the functionality they provide. `Makefiles` are historically related to [`Make`](https://en.wikipedia.org/wiki/Make_(software)) build or compilation procedures. They come along with some confusing syntax or capabilities which are relevant to that history, but I feel often are burdensome for developers to learn just for the sake of automating development tasks (especially if they don't involve `Make`, compilation, or building executables).

[`just`](https://github.com/casey/just) was created to address the needs we have here with a `justfile`, automating development tasks. It specifically addresses [some `Makefile` idiosyncrasies within the documentation](https://github.com/casey/just?tab=readme-ov-file#what-are-the-idiosyncrasies-of-make-that-just-avoids). In addition to those mentions, I think it's nice that the tool was easy to learn, easy to install, and is cross-compatible with multiple operating systems by default. There are also built-in functions like [`os()`](https://github.com/casey/just?tab=readme-ov-file#functions) to determine the operating system, which seems to have been a challenge at some point in this project's history (`bash` vs `zsh` default shells + `conda` environments). Generally I consider this to be an experiment in gardening 🧑‍🌾, if we find the tool is causing toil we should remove it and try something else!

0.download_data/run.sh
d33bs: Thanks good call, I think I'll leave this as-is because I'm not sure of any additional error handling we should perform at this time. Generally the subshell changes here came up as a result of [`shellcheck`'s SC2103](https://www.shellcheck.net/wiki/SC2103).

d33bs: This comment made me wonder if we should be careful about exceptions. As a result, I added `set -e` within the top of the file to help make sure we exit properly.

d33bs:
Thank you @MattsonCam for your review and comments! After making some changes and responding I'll now merge this in.

PR #37, opened 2024-10-31
Perform quality control with `coSMicQC` with UMAP analyses

d33bs:
This PR performs quality control through `coSMicQC` complemented by UMAP analyses.

Along the journey towards this I discovered `CytoDataFrame` has trouble accessing the images from JUMP so I opened wayscience/cytodataframe#15 and wayscience/cytodataframe#5 .

The current implementation is a draft using a small subset of the data to help elicit feedback and incorporate changes quickly.


0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
```diff
@@ -0,0 +1,698 @@
...
+# avoid displaying plot export warnings
+logging.getLogger("bokeh.io.export").setLevel(logging.ERROR)
+
+# set the plate name for use throughout the notebook
+example_plate = "BR00117006"
+
+
+# -
+
+# ## Define utility functions for use within this notebook
```
jenna-tomkinson, line 56: For the future when we plan on processing all of the plates with coSMicQC, would it make more sense to move this UMAP analysis as a separate notebook from coSMicQC processing?

I can see an argument to keep it because the UMAPs can provide important info for each plate, but across 51 plates, I can see it as a lot. 

Would love your feedback here!

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
```diff
@@ -0,0 +1,698 @@
...
+    "Image_FileName_CellOutlines",
+    "Image_FileName_NucleiOutlines",
+    "Image_FileName_OrigAGP",
+    "Image_FileName_OrigDNA",
+    "Image_FileName_OrigRNA",
+    "Cytoplasm_AreaShape_BoundingBoxMaximum_X",
+    "Cytoplasm_AreaShape_BoundingBoxMaximum_Y",
+    "Cytoplasm_AreaShape_BoundingBoxMinimum_X",
+    "Cytoplasm_AreaShape_BoundingBoxMinimum_Y",
+]
```
jenna-tomkinson, line 232: Why were these metadata columns manually selected? I can see the importance in some but very curious why boundingbox columns are included.

jenna-tomkinson: Also is "Image_FileName_NucleiOutlines" something found in the SQLite output or was it added during CytoTable?

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
```diff
@@ -0,0 +1,698 @@
...
+    rf"r{record['Image_Metadata_Row']:02d}c{record['Image_Metadata_Col']:02d}f{record['Image_Metadata_Site']:02d}p(\d{{2}})-.*\.tiff"  
+    for record in df_merged_single_cells[
+        [
+            "Metadata_Well",
+            "Image_Metadata_Row",
+            "Image_Metadata_Col",
+            "Image_Metadata_Site",
+        ]
+    ].to_dict(orient="records")
+}
```
jenna-tomkinson: If we want to use nuclei outlines, would we just change that here? I recommend using nuclei outlines as the default as we tend to find outlier segmentation using nuclei mainly.

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
```diff
@@ -0,0 +1,698 @@
...
+    "%",
+    "of",
+    outliers_counts.iloc[0],
+    "include erroneous outliers of some kind.",
+)
+
+# ## Prepare data for analysis with pycytominer
+
+# +
+parquet_sampled_with_outliers = f"./{example_plate}_sampled_with_outliers.parquet"
```
jenna-tomkinson, line 358: Where does this come from? I might have missed something being saved above.

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
```diff
@@ -0,0 +1,698 @@
...
+            "s3://cellpainting-gallery/cpg0000-jump-pilot/source_4"
+            "/workspace/metadata/external_metadata/"
+            "JUMP-Target-1_compound_metadata.tsv"
+        ),
+        sep="\t",
+    ),
+    left_on="broad_sample",
+    right_on="broad_sample",
+)
+df_platemap_and_metadata
```
jenna-tomkinson, line 448: This is very interesting! I do believe the metadata has been saved to the repository but this is a new way of getting a file that I never thought of. Consider using the metadata that has been added to this repo if possible, I could be wrong.

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
jenna-tomkinson, line 1: This is an awesome notebook! But I do think it is a bit long and includes a lot of stuff. Can we include this as an example or how should we proceed when we have to do this for all of the plates?

0.5.quality_control/images/umap_comparison_with_and_without_erroneous_outliers_BR00117006.png
jenna-tomkinson, line 1: Very interesting and weird shift! Love the plot!

0.5.quality_control/images/umap_erroneous_outliers_BR00117006.png
jenna-tomkinson, line 1: The outliers are very heterogeneous and spread out across the plot, very interesting. Could be that for this dataset that the outliers aren't noisy enough to impact the clusters, but not sure.

0.download_data/2.download_images.py
```diff
@@ -157,10 +157,13 @@ def download_jump_cpg000_images_from_s3(
     return "data/images"
 
 
+# -
+
+
```
jenna-tomkinson: Is this supposed to be blank?

0.download_data/2.download_images.py
```diff
@@ -157,10 +157,13 @@ def download_jump_cpg000_images_from_s3(
...
 # +
 # reference the file without reading it entirely
 target_file = pq.ParquetFile(
-    "../0.download_data/data/plates/BR00117006/BR00117006.parquet"
+    (parquet_path := "../0.download_data/data/plates/BR00117006/BR00117006.parquet")
```
jenna-tomkinson, line 163: What if we wanted to download the images for two different plates, say one per cell line?

environment.yml
```diff
@@ -34,7 +46,8 @@ dependencies:
   # used for formatting code within .ipynb files
   - conda-forge::jupyterlab_code_formatter
   - pip:
-      - git+https://github.com/cytomining/pycytominer.git@8e3c28d3b81efd2c241d4c792edfefaa46698115
+      - pycytominer
```
jenna-tomkinson: I have been told before that it is preferred to install pycytominer from conda-forge instead, so I would recommend making the change from pip.

jenna-tomkinson (approved):
This is a really awesome PR! 🎉 I think everything is a wonderful start and there isn't much to change, but I would love to hear your thoughts about the suggestions I have made in regards to the future of processing all of the data.

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
d33bs: Great question! Generally I feel that the UMAP generation and plotting are heavy lifts but can be important to iterate on within the notebook environment. That said, if we're looping through the work it might be beneficial to consider this as work which occurs outside of a notebook kernel and instead as "plain" Python process. I'd vote to keep a prototypical notebook around no matter what to help collectively remember what we did and why.

There's some early work here which includes function that help abstract visualizations - we could lean into this more with the Pycytominer and embedding generation work, supplying a local Python package within the project to help abstract things further (and keep the notebook length down). I'd urge that this is probably the most sustainable option if we think we might use this work again (the pattern seems to come up in many projects; profiles -> cosmicqc -> pycytominer -> umap embeddings -> umap plot).

Another mechanism might be parameterized notebooks via [Papermill](https://github.com/nteract/papermill). These might be harder to maintain as they rely on the notebook environment and are a bit more bespoke (don't occur as frequently within existing Way Lab projects).

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
d33bs: > Why were these metadata columns manually selected? I can see the importance in some but very curious why boundingbox columns are included.

Thanks for asking! These were manually selected to help identify the outlier or non-outlier rows when used with cosmicqc. The boundingbox columns are used by cytodataframe to help identify which portion of the original and outline images to use for displaying objects within the notebook. The bounding box columns are referenced automatically when present within the dataframe passed into cytodataframe ([see here for more](https://wayscience.github.io/CytoDataFrame/python-api.html#src.cytodataframe.frame.CytoDataFrame.get_bounding_box_from_data)).

> Also is "Image_FileName_NucleiOutlines" something found in the SQLite output or was it added during CytoTable?

These originate from the SQLite file. To my knowledge we do nothing more there to format the column name or add it in. I believe the naming of the column is specified within a CellProfiler pipeline step. I referenced these columns to help build the mapping between original and outline images.

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
d33bs: Yes, I that's correct, I recall now we discussed moving over to nuclei over cell outlines. I'll make that change now.

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
d33bs: Thanks for checking, this is a filename which is used to either read or create a file using the same name. We reference it here first as it will be used to create or read the file below this line.

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
d33bs: I think in the initial approach here I thought there might be data missing which we needed for Pycytominer work. I'll double check on this and follow up.

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
d33bs: Seeing this comment here made me think of the [thoughts I provided here](https://github.com/WayScience/JUMP-single-cell/pull/37#discussion_r1901136477). Revisiting this, I feel that this should be a prototype and we could lean into abstracting the repetitive work into a local Python package which would be performed when running this against dozens of other plates.

0.download_data/2.download_images.py
d33bs: Thanks - no, I'll remove from the notebook!

environment.yml
d33bs: Good point! Let's shift to the conda release.

0.download_data/2.download_images.py
d33bs: Good question; I'd recommend using `df_full = pd.read_parquet(parquet_path, columns=target_flattened_columns)` and the other work in a loop of some kind as seen below with the paths to those selected plates (without re-running the work to find the columns before that point here). Similar to other parts of this review and discussion, we could lean into further abstraction with a local Python package here if we think it could be more useful or reduce the code found within the notebook.

0.5.quality_control/0.jump_umap_analysis_with_cosmicqc.py
d33bs: I found that none of the files under the `reference_plate_data` dir include the `well_number` which is how we join the data to the profiles. I think specifically we need `well_position` and `control_type` in order to perform the Pycytominer annotation and normalization steps correctly. Going to leave things here as-is as a result.

d33bs:
Thanks @jenna-tomkinson ! After making some changes based on your comments I'll now merge this in.

Based on the discussion surrounding simplification or reuse of the notebook materials it sounds like it could make sense to abstract the capability to a local Python package (just my opinion, there are other paths here too!). That work could be performed in subsequent PR's. The benefit might be that we could more quickly perform the patterned work of profiles -> cosmicqc -> UMAP embeddings -> UMAP plots . Anything that is an existing function could be promoted into the local package, along with any other repetitive steps. Interested to get your and the teams thoughts.

PR #38, opened 2025-01-07
Convert download image notebook into Papermill notebook and add minor coSMicQC UMAP changes

d33bs:
This PR converts the download image notebook into a [Papermill](https://github.com/nteract/papermill) notebook which can be parameterized to download all images by plate ID. As part of this I changed the way that the image data are stored which incurred a change to the coSMicQC UMAP notebook. While I was making those changes I opted to add new plots which illustrate where in the UMAP the individual outlier threshold set members appear (and rerun the full notebook). This seems to have caused the full UMAP to change a bit.

I didn't run this on the full image data because I'm concerned about storage space on my local machine. I imagine this could be modified to take all plates as an argument via a loop of some kind in the `run.sh` of the download step.


0.5.quality_control/images/umap_comparison_with_and_without_erroneous_outliers_BR00117006.png
jenna-tomkinson, line 1: The original UMAP shape changed, are you using a random state to ensure reproducibility or did something else change? Super curious!

jenna-tomkinson (approved):
LGTM! I had one question about the UMAPs, but overall excited to work on this!

0.5.quality_control/images/umap_comparison_with_and_without_erroneous_outliers_BR00117006.png
d33bs: Thanks! I'll double check on this as I'm not sure.

d33bs: This seems to have been impacted by issues surrounding determinism and random_state within the UMAP package. I believe this issue might be associated. https://github.com/lmcinnes/umap/issues/1108 . I'm going to try and use the random state of 42 as commented by others (instead of 0) to see if that fixes anything. Prior to this I saw the same output images each time when running the notebook multiple times and during multiple Jupyter sessions. It could be that the number of processors or a change of machines has changed the outcome here as well - I'm not sure. In general, if you search for issues associated with "random_state" on the UMAP package repository many issues come up.

d33bs: After moving to using random state 42 with the UMAP library and running 2 times, clearing cached data each time, I found that the results are non-deterministic. My understanding is that this is related to the UMAP package and not the work here (as per the UMAP issue I cited). As a result, I'm going to resolve here and post to that issue that we are also seeing random_state challenges within that package.

d33bs:
Thanks @jenna-tomkinson ! I explored random_state with the UMAP package and found this is likely an error in operations with that package (it produces non-deterministic results seemingly without regard to random_state arguments). As a result, I'm going to merge this in now to move forward as best we can.

PR #39, opened 2025-02-20
Apply single cell QC and extract failed indices

jenna-tomkinson:
# Apply single-cell QC and extract failed indices

In this PR, coSMicQC processes the CytoTable JUMP pilot outputs and detects poor-quality segmentations. Indices for the failed single-cells are extracted as compressed CSVs for downstream processing.


1.process_data/nbconverted/single_cell_qc.py
d33bs, line 1: So cool that this is a papermill notebook which can be reused!

1.process_data/nbconverted/single_cell_qc.py
```diff
@@ -0,0 +1,277 @@
...
+
+# In[6]:
+
+
+# find irregular shaped nuclei using thresholds that maximizes
+# removing most technical outliers and minimizes good cells
+feature_thresholds = {
+    "Nuclei_AreaShape_FormFactor": -2.35,
+    "Nuclei_Intensity_MassDisplacement_DNA": 1.5
+}
```
d33bs: Should we add this into `cosmicqc`?

1.process_data/run_qc.sh
```diff
@@ -0,0 +1,26 @@
...
+plates=($(find "$PARENT_FOLDER" -mindepth 1 -maxdepth 1 -type d -exec basename {} \;))
+
+# Print the count of folders
+echo "Number of plates found: ${#plates[@]}"
+
+# Using papermill, run single cell quality control on all plates
+for plate in "${plates[@]}"; do
+    papermill \
+    single_cell_qc.ipynb \
+    single_cell_qc.ipynb \
```
d33bs, line 21: I like what you did here with the loop! This part of the arguments passed to papermill specifies the output notebook filename, which can log the results. It could make sense to change this to have a unique plate name in the output, meaning you'd be able to theoretically view any of the plate's `cytodataframe` cells within the output notebooks. This way you'd have a reference point later on, should we discover a need to do a deeper dive into any of those plates for QC reasons.

1.process_data/nbconverted/single_cell_qc.py
```diff
@@ -0,0 +1,277 @@
...
+
+# read only the metadata from parquet file
+parquet.ParquetFile(merged_single_cells).metadata
+
+
+# In[4]:
+
+
+# set a list of metadata columns for use throughout
+metadata_cols = [
```
d33bs, line 64: Consider including `TableNumber` also if it's available.

1.process_data/nbconverted/single_cell_qc.py
```diff
@@ -0,0 +1,277 @@
...
+# ### Save the original indices for failed single cells to compressed CSV files
+
+# In[11]:
+
+
+# Create a new dataframe for poor nuclei outliers
+poor_qc_indices = poor_nuclei_outliers[metadata_cols].copy()
+poor_qc_indices["original_index"] = poor_qc_indices.index  # Store original index
+poor_qc_indices = poor_qc_indices.reset_index(drop=True)  # Reset index
+poor_qc_indices["cqc.compactness_outlier"] = True
```
d33bs: Minor suggestion for greater clarity on what the boolean stipulates.
```suggestion
poor_qc_indices["cqc.compactness_outlier.is_outlier"] = True
```

1.process_data/nbconverted/single_cell_qc.py
```diff
@@ -0,0 +1,277 @@
...
+# ### Generate a new dataframe to save the original indices for filtering prior to further preprocessing
+
+# In[8]:
+
+
+# Create a new dataframe for failed QC indices
+failed_qc_indices = irregular_nuclei_outliers[metadata_cols].copy()
+failed_qc_indices["original_index"] = failed_qc_indices.index  # Store original index
+failed_qc_indices = failed_qc_indices.reset_index(drop=True)  # Reset index
+failed_qc_indices["cqc.formfactor_displacement_outlier"] = True
```
d33bs: Minor suggestion for clarity about the boolean value indicator (this also follows the `cosmicqc` convention).
```suggestion
failed_qc_indices["cqc.formfactor_displacement_outlier.is_outlier"] = True
```

1.process_data/nbconverted/single_cell_qc.py
```diff
@@ -0,0 +1,277 @@
...
+poor_nuclei_outliers_cdf.sample(n=5, random_state=0)
+
+
+# ### Save the original indices for failed single cells to compressed CSV files
+
+# In[11]:
+
+
+# Create a new dataframe for poor nuclei outliers
+poor_qc_indices = poor_nuclei_outliers[metadata_cols].copy()
```
d33bs: Consider using more specific words for these dataframes than `failed` or `poor`. Seeing them both one after another made me wonder what the difference might be (does `failed` entail `poor` or vice-versa, etc.). Maybe the name of the column outlier indicator column below could be used within the variable label of the DF as well.

1.process_data/nbconverted/single_cell_qc.py
```diff
@@ -0,0 +1,277 @@
...
+poor_qc_indices["original_index"] = poor_qc_indices.index  # Store original index
+poor_qc_indices = poor_qc_indices.reset_index(drop=True)  # Reset index
+poor_qc_indices["cqc.compactness_outlier"] = True
+
+# Merge both outlier dataframes on metadata columns and original_index
+failed_qc_indices = failed_qc_indices.merge(
+    poor_qc_indices,
+    on=[*metadata_cols, "original_index"],
+    how="outer"
+)
```
d33bs: Seeing how this plays out I noticed it might be possible to:
- Define a `yaml` file like the one found within `cosmicqc` [here](https://github.com/WayScience/coSMicQC/blob/main/src/cosmicqc/data/qc_nuclei_thresholds_default.yml) (which includes your threshold sets with the names you've specified manually). 
- Use that `yaml` file to run `cosmicqc.label_outliers(..., feature_thresholds_file=<yaml_file>, ...)`
- Export the file as you need below (also skipping the step to populate the data manually below).

Doing this might lower the complexity involved with the code more generally. It'd also prepare for the threshold sets to be added to `cosmicqc` :) .

1.process_data/nbconverted/single_cell_qc.py
```diff
@@ -0,0 +1,277 @@
...
+# Calculate percentage removed
+num_outliers = failed_qc_indices["original_index"].nunique()
+total_cells = len(df_merged_single_cells)
+percentage_removed = (num_outliers / total_cells) * 100
+
+# Sort by original index to maintain consistent order
+failed_qc_indices = failed_qc_indices.sort_values(by="original_index").reset_index(drop=True)
+
+# Save the indices for failed single-cells as a compressed CSV file
+failed_qc_indices.to_csv(Path(f"{qc_results_dir}/{plate_id}_failed_qc_indices.csv.gz"), compression="gzip")  # noqa: E501
```
d33bs: Consider using Parquet to help preserve data types and avoid ambiguous interpretations from CSV. These are compressed anyways, so we they'll appear like a Parquet file would. You could also take advantage of reading the entirety of the files as a single table more naturally, should you need it (a directory of Parquet files with the same schema may be treated as a single Parquet dataset).

1.process_data/run_qc.sh
```diff
@@ -0,0 +1,26 @@
...
+    -p plate_id $plate
+done
+
+# Convert notebook to script
+jupyter nbconvert --to=script --FilesWriter.build_directory=./nbconverted single_cell_qc.ipynb
```
d33bs, line 26: Consider using [`jupytext`](https://jupytext.readthedocs.io/en/latest/index.html) which automates this work and would remove the need for code duplication here and in other projects.

1.process_data/single_cell_qc.ipynb
d33bs, line 1: So. darn. cool! To see the nuclei in the `cytodataframes`!

d33bs (approved):
Good job - I left a few suggestions and overall thought this LGTM! Cheers on finding outliers using the tools we've built together within the lab on quite large datasets!

1.process_data/nbconverted/single_cell_qc.py
jenna-tomkinson: Thank you lots for the recommendation! It is my new go-to now since we normally have to process multiple plates! :tada:

1.process_data/nbconverted/single_cell_qc.py
jenna-tomkinson: Hmm, this is a good question! Maybe we can to something similar to CytoTable presets? 

My thought is that we can have a section in cosmicqc that has the presets that are associated to each publication that uses cosmicqc. So, instead of setting a "standard" preset, we can make a list of all the different uses of cosmicqc as examples for people to use.

I am going to add this as an issue into cosmicqc! Thanks for asking this!

jenna-tomkinson: Add as this issue: https://github.com/WayScience/coSMicQC/issues/198

1.process_data/run_qc.sh
jenna-tomkinson: Yes, I remember us discussing this in person! I will say I really enjoy this idea, but maybe not for this project. Since we do not have all of the images and outlines downloaded for all 51 plates, looking at the notebook for those plates would not yield much insight due to lack of images.

I think I want to consider this for other projects that have easier access to the images lol.

Thank you again for the suggestion!

1.process_data/nbconverted/single_cell_qc.py
jenna-tomkinson: I will add that back in! Do we know what TableNumber represents?

1.process_data/nbconverted/single_cell_qc.py
jenna-tomkinson: Great point! I will do that now!

1.process_data/nbconverted/single_cell_qc.py
jenna-tomkinson: Done and done!

1.process_data/nbconverted/single_cell_qc.py
jenna-tomkinson: This is a fair point, I think we should be consistent with "failed" in coSMicQC. I will update the name of this variable now!

1.process_data/nbconverted/single_cell_qc.py
jenna-tomkinson: This is an excellent comment! I am in total agreement, but I think it is out-of-scope for this PR. I think this is great for a refactor, and I will add this as an issue for the repo!

Thank you!

1.process_data/run_qc.sh
jenna-tomkinson: OOOO! Okay this would be a very fun refactor for this repository, I will add this as an issue for the project to consider because that would be much easier than having to run nbconvert.

1.process_data/nbconverted/single_cell_qc.py
jenna-tomkinson: I did have this thought originally, but since parquet files (as far as I know) can not be viewed in a viewer (if that makes sense), I could not confirm what I was outputting. So, this was the easiest for me to confirm. Since I know that my outputs will look correct, then I will change these to be parquets instead.

1.process_data/single_cell_qc.ipynb
jenna-tomkinson: YES! Thank you for all the hard work in coSMicQC and CytoDataFrame that made this possible. It is honestly a life saver now and MUST be the standard!

jenna-tomkinson:
@d33bs Thank you for the review! I am merging now!

PR #43, opened 2025-03-05
Normalize QC Single Cell Data

MattsonCam:
This pr normalizes the merged QC'd single cell data using index files. Any constructive feedback is welcome when reviewing.


1.process_data/normalize_jump_pilot.ipynb
jenna-tomkinson, line 1: Looks like this notebook isn't run, so recommend confirming that this notebook was ran and a script was created with nbconvert.

Other than that, the QC application looks like it should work! But, recommend adding a check that makes sure that the number of rows removed matches the number of rows in the failed indices file as a sanity check.

jenna-tomkinson (approved):
LGTM! Before merging, please make sure to include an nbconverted script. It looks like this notebook wasn't ran but this PR is for normalization being ran so I would double check that. Nice and simple PR, nice job!

1.process_data/normalize_jump_pilot.ipynb
MattsonCam: Sounds good, I probably should have mentioned I plan on running the script after this review (to phase out any bugs).

MattsonCam:
Hey @jenna-tomkinson, I think I found a bug (and fixed it). Requesting a re-review. I also included the nbconverted python file this time.

1.process_data/nbconverted/normalize_jump_pilot.py
```diff
@@ -202,10 +224,20 @@ def fill_dmso(df):
     # Merge the dataframes based on the broad_sample column, again there will not be matching DMSO broad samples for the CRISPR and ORF plates
     sc_df = pd.merge(sc_df, broad_mapdf, how="left", on=merge_col)
 
+    # Apply QC filtering using data indices
+    if qc_indices_manifests_path.exists():
+        for qc_plate_indices_file in qc_indices_manifests_path.iterdir():
+            if plate_name in qc_plate_indices_file.name:
+                plate_indicesdf = pd.read_parquet(qc_plate_indices_file)
+                sc_df = sc_df.loc[~plate_indicesdf["original_index"]]
```
jenna-tomkinson: This might not work as it might expect the values in original_index to be a `boolean` (per ChatGPT). Note that the `original_index` column contains the int of the row from the original data frame per plate that failed QC. 

I recommend double-checking this on a sub-sample from one plate before running. What I found that was recommended was the use:

```python
sc_df = sc_df.drop(plate_indicesdf["original_index"])
```

Based on the documentation, the default of the `drop` function is to drop based on axis=0 or the index.

jenna-tomkinson (changes requested):
I think that the code for the dropping of indices might not actually work. I recommend running a quick check on the subset of the data, I have included a recommendation for what might fix this potential bug.

1.process_data/nbconverted/normalize_jump_pilot.py
MattsonCam: Glad you caught that. I thought I changed it to something else. I will make the change now

MattsonCam:
Hey @jenna-tomkinson, I think I fixed the problem. Requesting another re-review

jenna-tomkinson (approved):
LGTM! 🎉

MattsonCam:
Thanks @jenna-tomkinson, merging now!

PR #44, opened 2025-03-07
Isolation Forest Data Feature Importances Utility

MattsonCam:
Provides an algorithm, as a util, for computing data Isolation Forest feature importances based on path lengths. Open to any constructive feedback when reviewing.

### DIFF
--- a/3.analyze_data/utils/isolation_forest_data_feature_importance.py
+++ b/3.analyze_data/utils/isolation_forest_data_feature_importance.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from typing import Optional, Sequence, Union
+from typing import Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -12,7 +12,7 @@ class IsoforestFeatureImportance:
 
     def __init__(
         self,
-        _estimators: Sequence[DecisionTreeRegressor],
+        _estimators: list[DecisionTreeRegressor],
         _morphology_data: pd.DataFrame,
     ):
         """
@@ -40,7 +40,7 @@ class IsoforestFeatureImportance:
         _sample_idx: int,
     ) -> dict[dict[str, float]]:
         """
-        Aggregates and then returns feature importances per tree.
+        Returns aggregated feature importances across trees for a given sample.
 
         Parameters
         _tree_obj: Object for traversing tree.
@@ -57,13 +57,8 @@ class IsoforestFeatureImportance:
             feature_idx = _tree_obj.feature[node_id]
 
             if feature_idx >= 0:  # Ignore leaf nodes (-2)
-                """
-                Samples are often easier to seperate with less data.
-                Therefore, earlier features have greater importances.
-                """
-                feature_importances[self._morphology_data.columns[feature_idx]].append(
-                    1 / (1 + depth)
-                )
+                # Indicates if a feature is present in a tree (1)
+                feature_importances[self._morphology_data.columns[feature_idx]].append(1)
 
             if morphology_features.iloc[feature_idx] <= _tree_obj.threshold[node_id]:
                 node_id = _tree_obj.children_left[node_id]
@@ -73,10 +68,10 @@ class IsoforestFeatureImportance:
             depth += 1
 
         """
-        Gives a balanced view of expected feature importance in the tree.
-        As opposed to the total contribution, which increases with the (same) features.
-        Since features are choosen randomly, dividing feature importances
-        by the depth weights each feature by how soon the data could be isolated.
+        Weights each feature, responsible for splitting the sample, equally
+        with the inverse depth of the sample's leaf node (terminating node).
+        This allows features to be more important with how soon each sample
+        is isolated across trees.
         """
 
         return {
@@ -86,8 +81,8 @@ class IsoforestFeatureImportance:
             }
         }
 
-    def compute_isoforest_feature_importances(self) -> pd.DataFrame:
-        # Computes and returns the importance for all trees and samples
+    def compute_isoforest_importances(self) -> pd.DataFrame:
+        # Computes and returns the aggregated importance for all features and samples (if they exist) using lazy parallelization.
 
         isotree_importances = Parallel(n_jobs=-1)(
             delayed(self.save_tree_feature_importances)(
@@ -100,7 +95,6 @@ class IsoforestFeatureImportance:
         )
 
         isoforest_importances = {}
-        unique_feats = set()
 
         sample_isoforest_importances = {
             sample: defaultdict(list)
@@ -120,13 +114,12 @@ class IsoforestFeatureImportance:
                 isoforest_importances[sample][feature] = sum(importances) / len(
                     importances
                 )
-                unique_feats.add(feature)
 
         self._isoforest_importances = pd.DataFrame(isoforest_importances).T
 
         return self._isoforest_importances
 
-    def get_isoforest_importances(
+    def get_filtered_isoforest_importances(
         self, _features: Union[str, list[str]]
     ) -> dict[str, float]:
         """
@@ -140,7 +133,7 @@ class IsoforestFeatureImportance:
         _features = [_features] if isinstance(_features, str) else _features
 
         if self._isoforest_importances is None:
-            filtered_morphology_data = self.compute_isoforest_feature_importances()
+            filtered_morphology_data = self.compute_isoforest_importances()
 
         else:
             filtered_morphology_data = self._isoforest_importances[_features].copy()
--- a/3.analyze_data/utils/isolation_forest_data_feature_importance.py
+++ b/3.analyze_data/utils/isolation_forest_data_feature_importance.py
@@ -1,0 +1,123 @@
+from collections import defaultdict
+from typing import Optional, Sequence, Union
+
+import numpy as np
+import pandas as pd
+from joblib import Parallel, delayed
+from sklearn.tree import DecisionTreeRegressor, _tree
+
+
+class IsoforestFeatureImportance:
+    def __init__(
+        self,
+        _estimators: Sequence[DecisionTreeRegressor],
+        _morphology_data: pd.DataFrame,
+    ):
+        """
+        Parameters
+        ----------
+        _estimators: Scikit-learn decision tree regressor estimators.
+        _morphology_data: Data with only numerical morphology data (without metadata).
+        """
+
+        self._estimators = _estimators
+        self._morphology_data = _morphology_data
+
+    def save_tree_feature_importances(
+        self,
+        _tree_obj: _tree.Tree,
+        _leaf_id: int,
+        _sample_idx: int,
+    ) -> dict[dict[str, float]]:
+        """
+        Aggregates and then returns feature importances per tree.
+
+        Parameters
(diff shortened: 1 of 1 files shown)


MattsonCam:
Considering not weighting features based on their order in a tree. Would prefer to hear thoughts on this

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+        # Averaging feature importances per sample and feature across trees
+        for sample, feature_importances in sample_isoforest_importances.items():
+            isoforest_importances[sample] = defaultdict(list)
+            for feature, importances in feature_importances.items():
+                isoforest_importances[sample][feature] = sum(importances) / len(
+                    importances
+                )
+                unique_feats.add(feature)
+
+        self._isoforest_importances = pd.DataFrame(isoforest_importances).T
```
MikeLippincott, line 118: Are the dfs transposed so that the importance become features of a sort?

MikeLippincott (approved):
LGTM. Clean code. Excited to see the implementation

3.analyze_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: I think you have the right idea. The columns are features and the rows are samples, so each (row, column) pair is the feature importance for that sample. This data can then be processed in different ways to determine feature importance between subsets of data

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+            _sample_idx: {
+                feature: (sum(importances) / len(importances)) / (depth + 1)
+                for feature, importances in feature_importances.items()
+            }
+        }
+
+    def compute_isoforest_feature_importances(self) -> pd.DataFrame:
+        # Computes and returns the importance for all trees and samples
+
+        isotree_importances = Parallel(n_jobs=-1)(
```
MattsonCam, line 87: May also be worth changing `n_jobs` here to be a variable that can be passed in

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def __init__(
+        self,
+        _estimators: Sequence[DecisionTreeRegressor],
+        _morphology_data: pd.DataFrame,
+    ):
+        """
+        Parameters
+        ----------
+        _estimators: Scikit-learn decision tree regressor estimators.
```
gwaybio, line 21: where does the `Sequence` type come in?

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+        self._estimators = _estimators
+        self._morphology_data = _morphology_data
+        self._isoforest_importances = None
+
+    @property
+    def isoforest_importances(self):
+        if self._isoforest_importances is None:
+            raise ValueError("isoforest_importances have not been computed")
+
+        return self._isoforest_importances
```
gwaybio, line 34: I'm a bit confused, but probably b/c I don't understand properties too well, but why is it returning itself?

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+        _leaf_id: int,
+        _sample_idx: int,
+    ) -> dict[dict[str, float]]:
+        """
+        Aggregates and then returns feature importances per tree.
+
+        Parameters
+        _tree_obj: Object for traversing tree.
+        _leaf_id: Leaf node for the sample in the tree.
+        _sample_idx: Sample location in the dataframe.
```
gwaybio, line 48: it might be helpful to clarify why this is important

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+
+        node_id = 0  # Start at the root node
+        depth = 0
+        feature_importances = defaultdict(list)
+        morphology_features = self._morphology_data.iloc[_sample_idx].copy()
+
+        while node_id != _leaf_id:
+            feature_idx = _tree_obj.feature[node_id]
+
+            if feature_idx >= 0:  # Ignore leaf nodes (-2)
```
gwaybio, line 59: what is `-2`?

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+                """
+                Samples are often easier to seperate with less data.
+                Therefore, earlier features have greater importances.
+                """
```
gwaybio: it feels a bit strange to see a docstring here - why not add this as a more traditional comment or move to the function docstring?

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+                node_id = _tree_obj.children_right[node_id]
+
+            depth += 1
+
+        """
+        Gives a balanced view of expected feature importance in the tree.
+        As opposed to the total contribution, which increases with the (same) features.
+        Since features are choosen randomly, dividing feature importances
+        by the depth weights each feature by how soon the data could be isolated.
+        """
```
gwaybio: An extra bit of clarity here would be helpful

1. What do you mean by "balanced view"?
2. When you say "how soon the data could be isolated", do you mean "how soon the feature is able to isolate a single sample in the tree"?

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+                """
+                Samples are often easier to seperate with less data.
+                Therefore, earlier features have greater importances.
+                """
+                feature_importances[self._morphology_data.columns[feature_idx]].append(
+                    1 / (1 + depth)
+                )
+
+            if morphology_features.iloc[feature_idx] <= _tree_obj.threshold[node_id]:
+                node_id = _tree_obj.children_left[node_id]
```
gwaybio, line 64: Does this traverse the tree in all directions?

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+        _sample_idx: Sample location in the dataframe.
+        """
+
+        node_id = 0  # Start at the root node
+        depth = 0
+        feature_importances = defaultdict(list)
+        morphology_features = self._morphology_data.iloc[_sample_idx].copy()
+
+        while node_id != _leaf_id:
+            feature_idx = _tree_obj.feature[node_id]
```
gwaybio, line 57: Is the feature_idx always going to be equivalent to the index in `self._morphology_data.columns`?

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+
+        return {
+            _sample_idx: {
+                feature: (sum(importances) / len(importances)) / (depth + 1)
+                for feature, importances in feature_importances.items()
+            }
+        }
+
+    def compute_isoforest_feature_importances(self) -> pd.DataFrame:
+        # Computes and returns the importance for all trees and samples
```
gwaybio: I think it would be helpful to describe the nature of how this function works (e.g., how does Parallel and delayed help this function)

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+
+        isoforest_importances = {}
+        unique_feats = set()
+
+        sample_isoforest_importances = {
+            sample: defaultdict(list)
+            for sample in range(self._morphology_data.shape[0])
+        }
+
+        # Converting data from list of tree dictionaries to dictionary of samples
```
gwaybio, line 104: This is a helpful comment!

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+        for isotree in isotree_importances:
+            for sample, feature_importances in isotree.items():
+                for feature, importance in feature_importances.items():
+                    sample_isoforest_importances[sample][feature].append(importance)
+
+        # Averaging feature importances per sample and feature across trees
```
gwaybio, line 110: I understand why it useful to calculate average feature importances across trees, but I don't understand why it's useful to calculate it per sample.

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+        for sample, feature_importances in sample_isoforest_importances.items():
+            isoforest_importances[sample] = defaultdict(list)
+            for feature, importances in feature_importances.items():
+                isoforest_importances[sample][feature] = sum(importances) / len(
+                    importances
+                )
+                unique_feats.add(feature)
```
gwaybio: is this doing anything?

3.analyze_data/utils/isolation_forest_data_feature_importance.py
gwaybio: looking below, it actually doesn't look like you average feature importances, but only calculate per sample?

3.analyze_data/utils/isolation_forest_data_feature_importance.py
gwaybio: This looks like a good candidate to include as a comment! (what are the rows and columns)

3.analyze_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -0,0 +1,148 @@
...
+
+        self._isoforest_importances = pd.DataFrame(isoforest_importances).T
+
+        return self._isoforest_importances
+
+    def get_isoforest_importances(
```
gwaybio:
```suggestion
    def get_isoforest_feature_importances(
```

gwaybio: (because it seems you're also calculating sample and feature importances

gwaybio (approved):
A few clarification questions, which I bet you can solve easily (by adding a comment, for example)

The biggest comment I have is my confusion about sample importances. Specifically, it seems like you calculate feature importance per sample, but aren't we interested in feature importances in general?

3.analyze_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: Good point, I think the documentation mentions specifically mentions "list of ExtraTreeRegressor instances". I will change it

MattsonCam:
Thanks for the review @gwaybio and @MikeLippincott, merging now!

PR #45, opened 2025-03-13
Revert "Isolation Forest Data Feature Importances Utility"

MattsonCam:
Reverts WayScience/JUMP-single-cell#44. Did not remove averaging of feature importances per tree as mentioned in comments


MattsonCam:
I will create another pr for this

PR #46, opened 2025-03-13
Changed feature importance aggregation per tree

MattsonCam:
This was discussed in the previous (pr #44 ) and wasn't implemented as was intended, so there is no need for another pr review.


MattsonCam:
Merging now!

PR #49, opened 2025-04-15
Compute single-cell anomaly data

MattsonCam:
This pr computes the anomaly data for each single-cell dataset. I have not tested this code yet and haven't used dask before, so any constructive feedback is welcome in the review.


2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -0,0 +1,82 @@
...
+from sklearn.ensemble import IsolationForest
+
+
+# ## Identify Anomaly Data
+
+# In[ ]:
+
+
+def compute_sc_outlier_data(
+    _scdf: pd.DataFrame,
```
d33bs: Double checking: is this a `pd.DataFrame` or a `dask.dataframe`?

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Compute single-cell outlier data.
+
+    Parameters
+    ----------
+    _scdf: Single cell profile data containing morphology features and metadata.
+    """
+
+    # Isolation forest reference:
+    # https://ieeexplore.ieee.org/document/4781136
```
d33bs: Nice reference! Would it make sense to start a [CITATION.cff](https://github.com/citation-file-format) file to help keep track of this and other references (using the `references` key)?

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -0,0 +1,82 @@
...
+
+# In[ ]:
+
+
+feat_cols = iso_forest.feature_names_in_
+meta_cols = [col for col in scddf.columns if "Metadata" in col]
+
+meta_dict = {col: scddf[col].dtype for col in meta_cols}
+meta_dict["Result_inlier"] = "i1"  # int8
+meta_dict["Result_anomaly_score"] = "f8"  # float64
```
d33bs: What do the comments besides these lines mean? Consider adding additional details about this just in case.

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
d33bs, line 1: Where does this file fit into the the "big picture" for how this is processed? Consider adding an entry to the `justfile` to help be explicit about how this is supposed to be executed. I mention this as a suggestion because it's easy to lose track of which steps are required before or after this one. Keeping the process/execution declarative helps to version this with the rest of the code - and also makes it clear for maintainers how this should work in practice.

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
d33bs, line 1: It looks like this notebook wasn't executed. I recognize you mentioned it hasn't yet been tested. Just the same, I wondered based on the context of the `.sh` file whether this could just be a Python module which is executed without a notebook. This would remove a step involved with converting the notebook to a Python file within the `.sh` file.

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -0,0 +1,82 @@
...
+import pandas as pd
+from sklearn.ensemble import IsolationForest
+
+
+# ## Identify Anomaly Data
+
+# In[ ]:
+
+
+def compute_sc_outlier_data(
```
d33bs: Consider writing a small test for this function to ensure it works the way you think it should. This could involve "promoting" the function to a Python module, where it might more easily be tested and imported.

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -0,0 +1,82 @@
...
+    _scdf: pd.DataFrame,
+) -> pd.DataFrame:
```
d33bs: Again, just double checking, is this a `pd.DataFrame`?

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -0,0 +1,82 @@
...
+    _scdf: Single cell profile data containing morphology features and metadata.
+    """
+
+    # Isolation forest reference:
+    # https://ieeexplore.ieee.org/document/4781136
+    isofor = IsolationForest(n_estimators=1_000, random_state=0, n_jobs=-1)
+    _scdf = _scdf.assign(Result_inlier=isofor.fit_predict(_scdf[feat_cols].copy()))
+    _scdf = _scdf[meta_cols].assign(
+        Result_anomaly_score=isofor.decision_function(_scdf[feat_cols].copy())
+    )
```
d33bs: Consider adding a comment about the importance of `copy()` here (without context I wasn't sure about why one might need to use `copy()` over a direct reference).

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
d33bs: This might be petty semantics, and if so, please ignore me! Would it be more accurate to call this `compute_sc_anomaly_score`? I wasn't sure if we wanted to be more specific about what an "outlier" is vs "anomaly". It could be important to distinguish given that `coSMicQC` uses the "outlier" language.

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
```diff
@@ -0,0 +1,40 @@
...
+
+conda init bash
+conda activate jump_sc
+
+# Get the root directory
+git_root=$(git rev-parse --show-toplevel)
+
+py_path="nbconverted"
+
+jupyter nbconvert --to python --output-dir="${py_path}/" *.ipynb
```
d33bs: Consider using `shellcheck` on this, which is a part of the pre-commit suite of tooling in this repo. A suggestion from that linter: consider using a relative path here to more safely specify the glob input (rule [SC2035](https://www.shellcheck.net/wiki/SC2035)).

```suggestion
jupyter nbconvert --to python --output-dir="${py_path}/" ./*.ipynb
```

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
```diff
@@ -0,0 +1,40 @@
...
+
+iso_forest_paths=f"${git_root}/2.evaluate_data/train_sc_anomalyze_models/isolation_forest_models"
```
d33bs: Consider removing this if it is not used.

d33bs (approved):

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: This should be a pandas dataframe, because sklearn isn't compatible with dask dataframes. Luckily dask uses pandas

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: I like the idea, I think this would provide a nicer structure. Right now, I think I'll stick with the one inline reference and consider expanding if more appear.

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: Good idea, I think it will make more sense if I add a comment about what the `meta` parameter means in the function call below these lines

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
MattsonCam: Forgot about that, good idea! Even though they will have to change the data paths in this code I think it will still be beneficial as a reference

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: That's true, and I think that would improve the organization. Much of the code in this repo has followed this pattern, and I think right now it's good enough. I'll keep it in mind for next time.

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
MattsonCam: I think I'll include other code in the justfile as well, so I'll probably create a new pr for this

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: Definitely will test it out

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: It should be, but I'll make sure

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: I was just being overly cautious. I'll remove it, since I don't think it is necessary

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: Probably a good idea to keep it consistent. An anomaly is just a type of outlier that is interesting or important, so this is computing outliers that aren't all strictly anomalies. Although, we are still computing anomalies.

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
MattsonCam: I can see how this might be unsafe in some programs, so I'll change it

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
MattsonCam: Good catch, one of the next lines should have an "s" ending (i.e. paths)

MattsonCam:
Can you give this another review @d33bs? I made many changes by deciding to not pursue the dask approach

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
```diff
@@ -10,11 +10,11 @@ git_root=$(git rev-parse --show-toplevel)
 
 py_path="nbconverted"
 
-jupyter nbconvert --to python --output-dir="${py_path}/" *.ipynb
+jupyter nbconvert --to python --output-dir="${py_path}/" ./*.ipynb
 
-iso_forest_paths=f"${git_root}/2.evaluate_data/train_sc_anomalyze_models/isolation_forest_models"
+iso_forest_paths="${git_root}/2.evaluate_data/train_sc_anomalyze_models/isolation_forest_models"
```
d33bs, line 15: Should this still be an f-string?

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
```diff
@@ -26,11 +26,19 @@ for plate_dir in "${plate_paths[@]}"; do
...
+        for file in "$plate_dir"/*.parquet; do
 
-        echo -e "\nSampling from $plate_dir"
+            if [ -f "$file" ]; then
 
-        /usr/bin/time -v python3 "$py_path/compute_sc_outlier_data.py" "$plate_dir" "$iso_forest_path"
+                iso_forest_path="$iso_forest_paths/$(basename "$plate_dir")_isolation_forest.joblib"
+
+
+                /usr/bin/time -v python3 "$py_path/compute_sc_anomaly_data.py" "$file" "$iso_forest_path" "sc_anomaly_data"
```
d33bs: Could `time` be located in another directory, depending on where this runs? Consider using a relative instead of absolute reference if possible.

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -10,73 +10,60 @@
...
-meta_dict["Result_inlier"] = "i1"  # int8
-meta_dict["Result_anomaly_score"] = "f8"  # float64
+scdf.sort_values(by="Result_anomaly_score", ascending=True, inplace=True)
 
-outlier_ddf = scddf.map_partitions(compute_sc_outlier_data, meta=meta_dict)
-outlier_ddf.to_parquet(
-    anomaly_data_path / sc_data_path.name / "_anomaly_data.parquet", write_index=False
+pq.write_to_dataset(
+    pa.Table.from_pandas(
+        scdf[
```
d33bs: Would it make sense to explore `pd.DataFrame.to_parquet(partition_cols=)` in context with this method? This might prevent a conversion outright to a PyArrow table.

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -10,73 +10,60 @@
...
 
-meta_dict = {col: scddf[col].dtype for col in meta_cols}
-meta_dict["Result_inlier"] = "i1"  # int8
-meta_dict["Result_anomaly_score"] = "f8"  # float64
+scdf.sort_values(by="Result_anomaly_score", ascending=True, inplace=True)
 
-outlier_ddf = scddf.map_partitions(compute_sc_outlier_data, meta=meta_dict)
-outlier_ddf.to_parquet(
-    anomaly_data_path / sc_data_path.name / "_anomaly_data.parquet", write_index=False
+pq.write_to_dataset(
```
d33bs: Consider making use of the `partition_cols` parameter to further split data by partitions (depending on what's best for the project).

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
d33bs: @MattsonCam - keep in mind one additional option for this which is slightly more Pythonic (though only because it involves the use of a `pyproject.toml`): [`poethepoet`](https://github.com/nat-n/poethepoet). That project enables similar capabilities but stores the logic in a `pyproject.toml`. `pyproject.toml` files can be useful for storing environment, configuration, and other Python project details.

d33bs (approved):

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
MattsonCam: I think I made the same mistake, this is bash and not python

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
MattsonCam: Not sure where this would be on non-linux systems. If I used a relative path I think I would have to go to root first, so if `time` is in a different directory I think the relative path would also be different.

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: That's true, I think it could avoid some overhead. I was writing parquet files this way to prevent overwriting of existing parquet files.

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
```diff
@@ -14,6 +14,8 @@ jupyter nbconvert --to python --output-dir="${py_path}/" ./*.ipynb
 
 iso_forest_paths="${git_root}/2.evaluate_data/train_sc_anomalyze_models/isolation_forest_models"
 
+anomaly_data_path="${git_root}/big_drive/sc_anomaly_data"
```
d33bs, line 17: Could it make sense to use Isilon / `bandicoot` for this so we may abstract the use across the team?

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -47,23 +46,14 @@
...
-                "Result_inlier",
-                "Result_anomaly_score",
-            ]
-        ]
-    ),
-    root_path=anomaly_data_path,
-)
+for i, batch in enumerate(pq_file.iter_batches(batch_size=220_000)):
+    pdf = batch.to_pandas()
+    meta_cols = [col for col in pdf.columns if "Metadata" in col]
```
d33bs, line 52: Instead of sniffing this out of each iteration, could this be gathered with a quick query ahead of time? I think there's a way to gather just the column names without referencing the record values through Parquet.

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -47,23 +46,14 @@
...
+    pdf = pdf.assign(Result_inlier=iso_forest.predict(pdf[feat_cols]))
+    pdf = pdf.assign(Result_anomaly_score=iso_forest.decision_function(pdf[feat_cols]))
```
d33bs, line 54: Double checking: does the model allow for accurate findings through iterations / pages / chunks of the dataset? Sometimes I think this is called "partial fitting/predicting/scoring" and while there are methods which enable it not all of them do.

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
```diff
@@ -47,23 +46,14 @@
...
-            meta_cols
-            + [
-                "Result_inlier",
-                "Result_anomaly_score",
-            ]
-        ]
-    ),
-    root_path=anomaly_data_path,
-)
+for i, batch in enumerate(pq_file.iter_batches(batch_size=220_000)):
```
d33bs, line 50: Consider adding a comment as to why you took this approach to help document why you took this route. I think I understand some of the background, but it could be easy to forget and fall back into the other patterns.

d33bs (approved):

2.evaluate_data/compute_sc_anomaly_data/compute_all_sc_anomaly_data.sh
MattsonCam: Maybe, I could have just run it on Alpine, but sometimes getting the environment to work can be painful

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: Thanks for checking, yes it does. This model is deterministic once trained

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: Good point, I only explained this to you in discord. I will add a comment here.

2.evaluate_data/compute_sc_anomaly_data/nbconverted/compute_sc_anomaly_data.py
MattsonCam: I think so, though I think this has a higher `development time : value` ratio

MattsonCam:
Thanks for all of the reviews @d33bs, merging now!

PR #50, opened 2025-05-09
Fix Feature Selection and Memory Bugs

MattsonCam:
This pr fixes two bugs. The first bug caused some columns that contain NaNs to be included when training anomalyze models. The second bug caused too much memory usage.

### DIFF
--- a/2.evaluate_data/train_sc_anomalyze_models/nbconverted/identify_anomalous_single_cells_fs.py
+++ b/2.evaluate_data/train_sc_anomalyze_models/nbconverted/identify_anomalous_single_cells_fs.py
@@ -9,6 +9,7 @@
 # In[ ]:
 
 
+import json
 import pathlib
 import sys
 
@@ -25,12 +26,21 @@ from sklearn.ensemble import IsolationForest
 # In[1]:
 
 
-plate_data_name = pathlib.Path(sys.argv[1]).name
+plate_data_path = pathlib.Path(sys.argv[1])
+plate_data_name = plate_data_path.name
+plate_data_dir_name = plate_data_path.parent.name
 sampled_plate_jump_data_path = sys.argv[2]
 
 sampled_platedf = pd.read_parquet(
     f"{sampled_plate_jump_data_path}/{plate_data_name}.parquet"
 )
+
+feature_columns_path = (
+    pathlib.Path(sys.argv[3]) / plate_data_dir_name / "feature_columns.json"
+).resolve(strict=True)
+
+with feature_columns_path.open("r") as feat_cols_obj:
+    feat_cols = json.load(feat_cols_obj)
 
 
 # ### Outputs
@@ -52,7 +62,7 @@ isoforest_path = pathlib.Path(
 
 
 meta_cols = [col for col in sampled_platedf.columns if "Metadata" in col]
-featdf = sampled_platedf.drop(columns=meta_cols).dropna(axis=1, how="any")
+featdf = sampled_platedf[feat_cols]
 
 # If 1_600 trees are trained with 256 samples per tree, then
 # 1_600 * 256 gives approximately the expected number of samples per tree.
--- a/2.evaluate_data/train_sc_anomalyze_models/nbconverted/identify_anomalous_single_cells_fs.py
+++ b/2.evaluate_data/train_sc_anomalyze_models/nbconverted/identify_anomalous_single_cells_fs.py
@@ -28,7 +28,6 @@ from sklearn.ensemble import IsolationForest
 
 plate_data_path = pathlib.Path(sys.argv[1])
 plate_data_name = plate_data_path.name
-plate_data_dir_name = plate_data_path.parent.name
 sampled_plate_jump_data_path = sys.argv[2]
 
 sampled_platedf = pd.read_parquet(
@@ -36,7 +35,7 @@ sampled_platedf = pd.read_parquet(
 )
 
 feature_columns_path = (
-    pathlib.Path(sys.argv[3]) / plate_data_dir_name / "feature_columns.json"
+    pathlib.Path(sys.argv[3]) / plate_data_name / "feature_columns.json"
 ).resolve(strict=True)
 
 with feature_columns_path.open("r") as feat_cols_obj:
@@ -48,7 +47,7 @@ with feature_columns_path.open("r") as feat_cols_obj:
 # In[ ]:
 
 
-isoforest_path = pathlib.Path("isolation_forest_models")
+isoforest_path = pathlib.Path(sys.argv[4])
 isoforest_path.mkdir(parents=True, exist_ok=True)
 
 isoforest_path = pathlib.Path(
--- a/2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
+++ b/2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
@@ -2,13 +2,14 @@
 # coding: utf-8
 
 # # Sample data from each well
-# Deterministically sample data from each plate dataset
+# Deterministically sample data from each plate dataset and get the intersection of all features not containing nans.
 
 # ### Import Libraries
 
 # In[ ]:
 
 
+import json
 import pathlib
 import sys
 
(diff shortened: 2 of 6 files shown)


MattsonCam:
I will also push the anomalyze models to this pr after they are finished training.

2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
```diff
@@ -40,6 +47,27 @@
 plate_data_path = sampled_plate_jump_path / f"{plate_path.parent.name}.parquet"
 
 
+# ## Save feature columns that don't contain NaNs
+
+# In[ ]:
+
+
+feat_cols = []
+non_na_cols = platedf.columns[platedf.notna().all()].tolist()
```
d33bs, line 56: Double checking: would this be able to find all `nan`-like values within the dataset? For example, would it grab `pd.NA`, `np.nan`, `None`, and `"nan"`? I bring this up only because I've seen these within other datasets so it could be important to make sure.

d33bs, line 56: Consider using a different variable label here (it could be confusing to read this later or to misuse the data on accident).

2.evaluate_data/train_sc_anomalyze_models/nbconverted/identify_anomalous_single_cells_fs.py
```diff
@@ -25,20 +26,28 @@
 # In[1]:
 
 
-plate_data_name = pathlib.Path(sys.argv[1]).name
+plate_data_path = pathlib.Path(sys.argv[1])
```
gwaybio, line 29: when using `sys.argv` I typically prefer, at the beginning of the notebook, assigning them to meaningful variable names (helps with readability)

2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
```diff
@@ -2,13 +2,14 @@
 # coding: utf-8
 
 # # Sample data from each well
-# Deterministically sample data from each plate dataset
+# Deterministically sample data from each plate dataset and get the intersection of all features not containing nans.
```
gwaybio, line 5: Are you getting nans b/c you're using the feature selected profiles? Does this impact the pre-trained isolation forest? (if it was trained with all features, i mean)

2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
```diff
@@ -40,6 +47,27 @@
...
+
+if feat_col_path.exists():
+    with feat_col_path.open("r") as feat_col_obj:
+        feat_cols = json.load(feat_col_obj)
+        feat_cols = list(set(non_na_cols) & set(feat_cols))
+
+else:
+    feat_cols = non_na_cols
+
+with feat_col_path.open("w") as feat_col_obj:
```
d33bs, line 67: Double checking: we want to truncate this file when we write to it, correct?

2.evaluate_data/utils/DeterministicSampling.py
```diff
@@ -21,34 +21,23 @@ def __init__(
         self._cell_id_columns = _cell_id_columns
         self._divisor = 10_000
 
-    """
-    _plate_column, _well_column and _cell_id_columns store column names, and must be present in _platedf.
-    _cell_id_columns will be used with _plate_column and _well_column to uniquely identify each cell.
-    For example, _cell_id_columns in some projects could be ["Metadata_Site", "Metadata_ObjectNumber"]
-    """
```
d33bs: This looked like a good comment - how come we removed it?

2.evaluate_data/utils/DeterministicSampling.py
```diff
@@ -21,34 +21,23 @@ def __init__(
...
+        # Vectorized string join for hashing
+        combined_strs = self._platedf[hash_cols].astype(str).agg("".join, axis=1)
 
-        self._platedf["Metadata_farmhash"] = self._platedf.apply(
-            lambda row: Fingerprint64("".join(str(row[col]) for col in hash_cols))
-            % self._divisor,
-            axis=1,
+        # Vectorized hash mapping
+        self._platedf["Metadata_farmhash"] = combined_strs.map(
+            lambda x: Fingerprint64(x) % self._divisor
```
d33bs, line 44: I took a moment to look up [`farmhash`](https://github.com/google/farmhash) where `Fingerprint64` is sourced from. It looks like it's an archived project as of 2023. It could be important to use an alternative implementation of this from a package which is actively maintained to avoid software decay.

2.evaluate_data/utils/DeterministicSampling.py
```diff
@@ -57,26 +46,18 @@ def sample_plate_deterministically(self, _sample_strategy: str = "well_sampling"
...
+        return pd.concat(
+            (
+                group_df.loc[
+                    group_df["Metadata_farmhash"]
+                    < (samples_per_group / len(group_df)) * self._divisor
+                ]
+                for _, group_df in grouped
+            ),
+            ignore_index=True,
+        )
```
d33bs, line 71: This looks like complex logic; consider adding a comment or two about how the grouping works and what the product of this comprehension will be.

d33bs (approved):

2.evaluate_data/utils/DeterministicSampling.py
```diff
@@ -21,34 +21,23 @@ def __init__(
...
-    """
-
     @property
     def platedf(self):
         return self._platedf.copy()
 
     def __hash_data(self):
-        hash_cols = [self._plate_column, self._well_column]
-        self._group_cols = hash_cols.copy()
+        hash_cols = [self._plate_column, self._well_column] + self._cell_id_columns
```
gwaybio, line 37: I do not understand what is going on here! Please consider adding a comment

gwaybio (approved):
I made a few comments below. In general, it's unclear to me why `identify` and `sample` are in two different notebooks.

2.evaluate_data/utils/DeterministicSampling.py
MattsonCam: Ya, from what I understand Google created the farmhash. I use [this implementation](https://github.com/escherba/python-cityhash), which they claim is hardware-independent

2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
MattsonCam: When I checked, it doesn't seem that None and "nan" are in these numeric columns. I think pandas had to fill in some missing values, which is why there are nans in this case

MattsonCam: Thanks, I'll consider it

2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
MattsonCam: Correct

2.evaluate_data/utils/DeterministicSampling.py
MattsonCam: I think that was a mistake, good catch

2.evaluate_data/utils/DeterministicSampling.py
MattsonCam: Good idea

2.evaluate_data/utils/DeterministicSampling.py
MattsonCam: Sounds good

2.evaluate_data/train_sc_anomalyze_models/nbconverted/identify_anomalous_single_cells_fs.py
MattsonCam: Definitely agree

2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
MattsonCam: Yes it does, although fortunately, I think it samples quicker now

MattsonCam:
Thanks for the reviews, merging now!

PR #51, opened 2025-05-12
Updated Training Size and Models

MattsonCam:
This pr updates the training size so as to not to exceed the memory threshold while training anomalyze. I am also including the trained models in this pr.


2.evaluate_data/train_sc_anomalyze_models/sample_data_and_train_anomalyze.sh
```diff
@@ -24,10 +24,10 @@ jupyter nbconvert --to python --output-dir="${py_path}/" *.ipynb
 
 # Get the normalized data path (with multiple plates)
 plate_paths=(
+    "${git_root}/big_drive/normalized_sc_data"
```
d33bs, line 27: Could there be a way to generalize this path so it's abstracted from the system? I imagine `big_drive` is the name of a local directory and wondered how it could be generalized for use on any system.

2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
```diff
@@ -77,9 +77,9 @@
     _platedf=platedf,
     # Number of samples per plate needed to train the JUMP isolation forests
     # See identify_anomalous_single_cells_fs.py for more details
-    # Only 8_000 are needed, however each sampling will likely not be exactly 8_000 samples.
+    # Only 4_000 are needed, however each sampling will likely not be exactly 4_000 samples.
```
d33bs: Consider adding additional context here if possible (for ex. "why 4000?" and "why not exactly 4000?").

2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
```diff
@@ -77,9 +77,9 @@
...
     # To ensure a sufficient sample size I increased the threshold.
-    _samples_per_plate=8_100,
+    _samples_per_plate=4_100,
```
d33bs, line 83: Is the comment above still applicable (about increasing the threshold)? If not, consider revising.

2.evaluate_data/train_sc_anomalyze_models/isolation_forest_models/feature_selected_sc_data_isolation_forest.joblib
d33bs, line 1: Could it make sense to use ONNX models here instead of `joblib`? Maybe [`sklearn-onnx`](https://onnx.ai/sklearn-onnx/) and [`onnxruntime`](https://onnxruntime.ai/docs/get-started/with-python.html) could be useful here.

d33bs (approved):

2.evaluate_data/train_sc_anomalyze_models/sample_data_and_train_anomalyze.sh
MattsonCam: I don't think it could be unless you somehow agreed to some shared standard, such as the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/). Although I don't think Windows uses the standard, and I think you would have to set them in macos. Also, I don't think there would be a good XDG environment variable to account for the heterogeneity of storage on a system, so I think the user knows best for their system. I wouldn't consider this to be a tool in the same way as pairwise compare or pycytominer, and more so as a means to show our approach and provide the code if the user wants to reproduce the analysis.

2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
MattsonCam: Sounds good, I'll briefly mention hashing here

2.evaluate_data/train_sc_anomalyze_models/nbconverted/sample_anomalous_single_cells_fs.py
MattsonCam: The comment above is still applicable

2.evaluate_data/train_sc_anomalyze_models/isolation_forest_models/feature_selected_sc_data_isolation_forest.joblib
MattsonCam: I am a fan of ONNX models, however I decided to use joblib because it seemed like a lab standard when I first joined. Although, maybe it would be worth considering the idea again

MattsonCam:
Thanks @d33bs, merging now!

PR #53, opened 2025-05-20
Improve feature importance algorithm

MattsonCam:
Improves on the previous approach for determining feature importance by deriving the feature importance scores from the anomaly scores equation. This will later be used to assess the features contributing the most to identifying outliers.


2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -8,12 +8,19 @@
 
 
 class IsoforestFeatureImportance:
-    # Computes Isolation Forest (Scikit-learn) feature importances from a subset of data.
+    """
+    Computes Isolation Forest (Scikit-learn) feature importances from a subset of data.
+    Feature importances were derived from the Isolation Forest paper.
+    See https://doi.org/10.1109/ICDM.2008.17 for more details.
+    This approach assumes each feature between a sample's leaf node and root node
+    is equally important in isolating that sample for a given isolation tree estimator.
```
d33bs, line 16: I love this documentation! Would it make sense to add the DOI (and related metadata) to a [`CITATION.cff`](https://github.com/citation-file-format/citation-file-format)?

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -33,6 +43,18 @@ def isoforest_importances(self):
 
         return self._isoforest_importances
 
+    def _compute_norm_factor(self, _num_features_per_forest: int):
```
d33bs: Consider adding a test for this within the repo somehow. For example, one could use `pytest`, or if it's not too troublesome, [`doctest`](https://docs.python.org/3/library/doctest.html).

d33bs: Consider adding a type for `self`.

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -33,6 +43,18 @@ def isoforest_importances(self):
...
+        """
+        Used to compute the anomaly score in an isolation forest.
```
d33bs, line 48: Consider adding the args and return for this docstring.

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -33,6 +43,18 @@ def isoforest_importances(self):
...
+        """
+
+        harmonic_approx = np.log(_num_features_per_forest) + np.euler_gamma
```
d33bs: How does `harmonic` apply here? Consider adding a short comment about "why" or "what" we're doing here.

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -67,24 +89,16 @@ def save_tree_feature_importances(
 
             depth += 1
 
-        """
-        Weights each feature, responsible for splitting the sample, equally
-        with the inverse depth of the sample's leaf node (terminating node).
-        This allows features to be more important with how soon each sample
-        is isolated across trees.
-        """
```
d33bs: Is this no longer true? Would it make sense to update it instead of completely removing?

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -67,24 +89,16 @@ def save_tree_feature_importances(
...
+                for feature, count in num_features.items()
             }
-        }
 
     def compute_isoforest_importances(self) -> pd.DataFrame:
-        # Computes and returns the aggregated importance for all features and samples (if they exist) using lazy parallelization.
+        # Computes feature importances for all features and samples (if they exist) using lazy parallelization.
 
-        isotree_importances = Parallel(n_jobs=-1)(
+        isotree_sample_importances = Parallel(n_jobs=-1)(
```
d33bs, line 101: Nice use of `joblib.Parallel`.

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -94,25 +108,38 @@ def compute_isoforest_importances(self) -> pd.DataFrame:
...
-            isoforest_importances[sample] = defaultdict(list)
-            for feature, importances in feature_importances.items():
-                isoforest_importances[sample][feature] = sum(importances) / len(
-                    importances
+        for isotree_sample in isotree_sample_importances:
+            for sample, sample_isotree_feature_counts in isotree_sample.items():
+                for feature, depths in sample_isotree_feature_counts.items():
+                    sample_isoforest_importances[sample][feature].extend(depths)
+
+        isoforest_importances = defaultdict(lambda: dict)
```
d33bs, line 125: Double checking: do you want to store a type reference to `dict` here or use a default of `{}` for values?

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -8,12 +8,19 @@
...
+    See https://doi.org/10.1109/ICDM.2008.17 for more details.
+    This approach assumes each feature between a sample's leaf node and root node
+    is equally important in isolating that sample for a given isolation tree estimator.
+    """
 
     def __init__(
         self,
         _estimators: list[DecisionTreeRegressor],
         _morphology_data: pd.DataFrame,
+        _num_train_samples_per_tree: int,
```
d33bs: Double checking on the use of the privacy underscore here. Consider dropping the underscores for initializer parameters we intend to receive from the outside, setting the attributes to the underscored values (which appear private). Otherwise, we might reason that the parameters seem like they're private (like we're not supposed to set them).

d33bs (approved):

MattsonCam:
Thanks for the review @d33bs. I've learned this code will take too long, so I plan to include sampling in this pr. Unfortunately, I think the bottleneck occurs when traversing the tree, which seems like the only feasible way of determining feature importance

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: Thanks! I think so, once we are ready to cite it

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: This could be useful, but I think it would have a higher value:time ratio if it were a tool for multiple analyses like pairwise compare. Right now I think it will only be used for the JUMP and durbin lab data

MattsonCam: Good to know, had no idea this was possible

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: Good point, forgot to include the return type hint

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: This is more of a technicality of how anomaly scores are calculated, where the harmonic number can be used to estimate the average path length of an unsucessful search in a BST, which resembles isolation forest trees. I think it would be better to read the above cited literature to understand this aspect

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: Correct, there is no need for this. I might rename the function entirely, because I am just recording the depths for each of the features here

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: Just the later

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: That's a good point, I'm in the habit of doing this to avoid name conflicts.

MattsonCam:
Thanks @d33bs, merging now!

PR #55, opened 2025-07-23
Fix Anomalyze Bug

MattsonCam:
This pr fixes a bug in anomalyze causing the data to not be stored correctly.

### DIFF
--- a/2.evaluate_data/utils/isolation_forest_data_feature_importance.py
+++ b/2.evaluate_data/utils/isolation_forest_data_feature_importance.py
@@ -89,11 +89,10 @@ class IsoforestFeatureImportance:
 
             depth += 1
 
-        if node_id != _leaf_id:
-            return {
-                _sample_idx: {feature: [depth] * count}
-                for feature, count in num_features.items()
-            }
+        return {
+            _sample_idx: {feature: [depth] * count}
+            for feature, count in num_features.items()
+        }
 
     def compute_isoforest_importances(self) -> pd.DataFrame:
         # Computes feature importances for all features and samples (if they exist) using lazy parallelization.
@@ -125,7 +124,7 @@ class IsoforestFeatureImportance:
                 for feature, depths in sample_isotree_feature_counts.items():
                     sample_isoforest_importances[sample][feature].extend(depths)
 
-        isoforest_importances = defaultdict(lambda: dict)
+        isoforest_importances = defaultdict(dict)
 
         # Computes the feature importance across all trees that
         # split using the corresponding feature.
--- a/2.evaluate_data/compute_sc_anomalyze_data/compute_feature_importance_by_plate.ipynb
+++ b/2.evaluate_data/compute_sc_anomalyze_data/compute_feature_importance_by_plate.ipynb
@@ -80,8 +80,8 @@
       },
       "source": [
         "# Number of samples to compute feature importances for in each category (anomalous or inlier)\n",
-        "num_control_samples = 100\n",
-        "num_anomalous_samples = 100\n",
+        "num_control_samples = 300\n",
+        "num_anomalous_samples = 300\n",
         "\n",
         "merge_cols = [\n",
         "    \"Metadata_Plate\",\n",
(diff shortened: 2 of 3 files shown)


2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -89,11 +89,10 @@ def save_tree_feature_depths(
 
             depth += 1
 
-        if node_id != _leaf_id:
-            return {
-                _sample_idx: {feature: [depth] * count}
-                for feature, count in num_features.items()
-            }
+        return {
```
wli51, line 92: Correct me if I am wrong, this seems to be fixing the method unexpectedly returning None when a single node tree is encountered. If that is the case this looks good to me. This fix does seem to have a subtle down-stream impact in `compute_isoforest_importances`, please see comment down.

wli51: A minor suggestion, perhaps the return type hint of `save_tree_feature_depths`, currently `-> dict[dict[str, float]]`, should be updated to reflect the actual return dict structure.

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -125,7 +124,7 @@ def compute_isoforest_importances(self) -> pd.DataFrame:
                 for feature, depths in sample_isotree_feature_counts.items():
                     sample_isoforest_importances[sample][feature].extend(depths)
 
-        isoforest_importances = defaultdict(lambda: dict)
+        isoforest_importances = defaultdict(dict)
```
wli51, line 127: LGTM

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -125,7 +124,7 @@ def compute_isoforest_importances(self) -> pd.DataFrame:
                 for feature, depths in sample_isotree_feature_counts.items():
```
wli51, line 124: This comment is referring to lines 113-117:

```
isotree_sample_importances = [
            tree_sample
            for tree_sample in isotree_sample_importances
            if tree_sample is not None
]
```
seems to be for handling `None` returns from `save_tree_feature_depths`, which should happen no more. Is there any other tricky edge cases we are trying to catch here? If not maybe consider removing this?

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -89,11 +89,10 @@ def save_tree_feature_depths(
 
             depth += 1
 
-        if node_id != _leaf_id:
-            return {
-                _sample_idx: {feature: [depth] * count}
-                for feature, count in num_features.items()
-            }
+        return {
+            _sample_idx: {feature: [depth] * count}
```
wli51, line 93: out of curiosity by the design of this return are you thinking of an alternative to return features with their actual depth at split vs. the current `[depth] * count`?

wli51 (approved):
Great job! Please see the comments for minor suggestions.

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: > Correct me if I am wrong, this seems to be fixing the method unexpectedly returning None when a single node tree is encountered. If that is the case this looks good to me. This fix does seem to have a subtle down-stream impact in `compute_isoforest_importances`, please see comment down.

Each tree will always have leaf nodes (otherwise the isolation tree wouldn't exist). Therefore, all of the variables in this return should not be None. When a leaf node is reached for a sample then we no longer need to traverse the tree because `node_id == _leaf_id`. Unfortunately, the previous bug would have prevented this result from being returned because the if statement will always be false, i.e. `node_id != _leaf_id`

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: True, I don't think this makes a difference. Although, maybe there is some edge case I'm not considering

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: Good question, this is done in parallel, so it should be returning features with their actual depth at split

MattsonCam:
Thanks @wli51, merging now!

PR #57, opened 2025-11-06
Save Metadata with Anomaly Feature Importances

MattsonCam:
Saves the metadata per sample with the anomaly feature importances.


2.evaluate_data/compute_sc_anomalyze_data/nbconverted/compute_sc_feature_importance_data.py
```diff
@@ -66,13 +66,13 @@
 # In[ ]:
```
axiomcura, line 67: Consider rerunning your notebook. The cells do not seem to be executed

2.evaluate_data/compute_sc_anomalyze_data/nbconverted/compute_sc_feature_importance_data.py
```diff
@@ -66,13 +66,13 @@
...
 
 
-feature_importances_path = root_dir / "big_drive/sc_feature_importances"
+feature_importances_path = root_dir / "big_drive/sc_anomaly_feature_importances"
```
axiomcura: This is not a suggestion but a heads up! This might be a reproducibility issue if someone were to run this on their local machine as "big_drive" may not be part of their pathing.

2.evaluate_data/compute_sc_anomalyze_data/nbconverted/compute_sc_feature_importance_data.py
```diff
@@ -128,7 +130,7 @@
                     str(anomalyze_models_path / anomaly_model_name),
                     str(treatment_feature_importances_path),
                     str(plate_mapping_path),
-                    str("compute_feature_importance_by_plate.py"),
+                    str("nbconverted/compute_feature_importance_by_plate.py"),
```
axiomcura: is the `str()` necessary here? Since what you are providing are strings already. It could be I might be missing something here.

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
```diff
@@ -73,7 +73,7 @@ def save_tree_feature_depths(
         node_id = 0  # Start at the root node
         depth = 0
         num_features = defaultdict(int)
-        morphology_features = self._morphology_data.iloc[_sample_idx].copy()
+        morphology_features = self._morphology_data.loc[_sample_idx].copy()
```
axiomcura: I do not think it is necessary to use the `copy()` method here. Looking at the rest of code, it seem that you are only reading the values not modifying of the data. I could be wrong!

axiomcura (approved):
LGTM! left some comments!

2.evaluate_data/compute_sc_anomalyze_data/nbconverted/compute_sc_feature_importance_data.py
MattsonCam: I'm running these from a bash script, so they won't be executed

2.evaluate_data/compute_sc_anomalyze_data/nbconverted/compute_sc_feature_importance_data.py
MattsonCam: Haha, I don't think so, I will change this

2.evaluate_data/compute_sc_anomalyze_data/nbconverted/compute_sc_feature_importance_data.py
MattsonCam: Good point, I added a comment so that they know to change the storage path based on where they are keeping the data on their filesystem

2.evaluate_data/utils/isolation_forest_data_feature_importance.py
MattsonCam: Maybe that will help a bit, thanks!

MattsonCam:
Thanks @axiomcura, merging now!

PR #58, opened 2025-12-20
Added compound moa

axiomcura:
This PR adds a metadata file with compound mechanism-of-action (MoA) information for the compound drug screen, sourced from the Drug Repurposing Hub ([link](https://repo-hub.broadinstitute.org/repurposing#download-data)￼). Note that we use the version generated on 07/09/2018; later versions may add or remove compounds, which would affect dataset reproducibility. The exact file used is available here: https://s3.amazonaws.com/data.clue.io/repurposing/downloads/repurposing_drugs_20180907.txt


MattsonCam (approved):
LGTM, it's good you were able to find this. Not sure exactly why they would remove compounds, but you know better than I do about this. In my opinion, it would be better to just keep them and add another column like "not_approved" or "date_included".

MattsonCam:
Since @axiomcura doesn't have permission to merge this pr, he gave me the okay to merge it.

PR #63, opened 2026-02-20
Compare Anomaly Scores based on Aggregation Order

MattsonCam:
Compares the anomaly scores of treatments based on the aggregation order. The two ways the anomaly scores can be aggregated are:
- Aggregated Profile: Aggregated single-cell profiles to treatment level -> Trained with anomalyze to compute anomaly scores
- Aggregated Anomaly Score: Trained with anomalyze to compute anomaly scores -> Aggregated single-cell anomaly scores to treatment level


3.analyze_data/compare_aggregate_single_cell_anomalies/figures/anomaly_comparisons.png
d33bs, line 1: Consider using some kind of visual depiction of "weighted overlap" here through binning or some other means. That is, the more points overlapping one another should be more pronounced, showing where the density is (because after full saturation, which happens quickly with opacity only, we can't tell exactly what the difference is).

Sidenote: this reminds me of Guile from Street Fighter's kick "slash".

Image

3.analyze_data/compare_aggregate_single_cell_anomalies/nbconverted/compare_aggregate_single_cell_anomalies.py
```diff
@@ -0,0 +1,142 @@
...
+# ## Inputs
+
+# In[2]:
+
+
+root_dir = pathlib.Path("../../").resolve(strict=True)
+
+agg_anomdf = pd.read_parquet(
+    root_dir
+    / "big_drive/aggregated_anomaly_data/aggregated_treatment_anomaly_data.parquet"
```
d33bs, line 30: Is there a way we could abstract `"big_drive"` for anyone on the team? Consider finding a way with the code here to do that and also a proposed method for the lab to use with regards to local, non-bandicoot filesystem usage.

3.analyze_data/compare_aggregate_single_cell_anomalies/nbconverted/compare_aggregate_single_cell_anomalies.py
```diff
@@ -0,0 +1,142 @@
...
+
+agg_platedf = []
+
+for plate_path in sc_anom_paths.iterdir():
+
+    treatment_col_name = "Metadata_pert_iname"
+    platedf = pd.concat(
+        [
+            pd.read_parquet(data_path)
+            for data_path in list(plate_path.rglob("*.parquet"))
```
d33bs: Could you iterate the generator instead of casting to list here? Also, do you need `rglob` for recursive or could you use `glob` instead?

3.analyze_data/compare_aggregate_single_cell_anomalies/nbconverted/compare_aggregate_single_cell_anomalies.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+
+    result_cols = platedf.columns[platedf.columns.str.contains("Result")]
+
+    agg_mapping = dict.fromkeys(result_cols, "mean")
+    agg_mapping |= {"Metadata_Plate": "first"}
+
+    # Distinguish between Compound data and (CRISPR and ORF data)
+    if not np.any(platedf.columns.str.contains(treatment_col_name)):
+        treatment_col_name = "Metadata_gene"
```
d33bs, line 72: Instead of changing around the use of this variable label, consider using the value directly where needed.

d33bs (approved):

3.analyze_data/compare_aggregate_single_cell_anomalies/figures/anomaly_comparisons.png
MattsonCam: Good idea, I think this is usually a good idea. In this figure the points don't seem to align with the reference line though, so I might hold off on this.

It does, a blast from the past. I could also see an upper-cut here

3.analyze_data/compare_aggregate_single_cell_anomalies/nbconverted/compare_aggregate_single_cell_anomalies.py
MattsonCam: Ya, I'm not a fan of this either, but the other code uses this convention so I think I'll grit and bear through it. I think a future solution could be to have the user set an environment variable for each repo, and reference that in the code

3.analyze_data/compare_aggregate_single_cell_anomalies/nbconverted/compare_aggregate_single_cell_anomalies.py
MattsonCam: Good point, I'll change this to not use casting.

MattsonCam: The rglob here is necessary because there are multiple parquet files per plate

3.analyze_data/compare_aggregate_single_cell_anomalies/nbconverted/compare_aggregate_single_cell_anomalies.py
MattsonCam: This is the tricky part, because depending on the treatment type, the dataframe may have a different column name, so I'm checking if one of the columns is present, and if not, I change the value of the column's name

MattsonCam:
Thanks @d33bs , merging now!

PR #64, opened 2026-03-14
Image Data Downloading and Processing

MattsonCam:
In this pr the JUMP image metadata is downloaded and processed. Then the image metadata can be filtered to download the desired images from aws S3 using a python utility. 

There is an example of how to do this, but I didn't include it in this pr yet. When you review this pr, please let me know where you think this example would fit best. Similarly, I named the utility `4.download_images_from_metadata.py` for now, but I'd appreciate any feedback on where you think this utility would belong. It will likely be used for both the anomalyze and nuclear speckle prediction projects in the future.

Also, I think I am processing the image metadata correctly, however please let me know if I'm not.


0.download_data/nbconverted/3.download_image_metadata.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+# ## Download and Concatenate Metadata
+
+# In[3]:
+
+
+storage_options = {"anon": True}
+
+fs = s3fs.S3FileSystem(anon=True)
```
d33bs: Consider adding short docs here why this was needed.

0.download_data/2.download_images.py
d33bs, line 1: Double checking: should this have been removed?

0.download_data/nbconverted/2.download_image_metadata.py
```diff
@@ -0,0 +1,100 @@
...
+    meta_plate_img_df = pd.read_csv(s3_url, storage_options=storage_options)
+    meta_plate_img_df["source_plate"] = plate
+    meta_plate_img_df["source_s3_path"] = s3_url
+    meta_img_df.append(meta_plate_img_df)
+
+meta_img_df = pd.concat(meta_img_df, axis=0, ignore_index=True)
+
+
+# ## Format Metadata
+# Format the metadata to use better channel names and specify one url per line
```
d33bs, line 59: Better how? Consider defining this a bit more.

0.download_data/utils/download_images_from_metadata.py
d33bs, line 1: I was confused to see this file here but the `3.#` file under `nbconverted`. Consider keeping consistent one where `.py` files go.

d33bs, line 1: If I'm understanding right, we removed the original image download code and built this new code to do some of that work, in addition to what was needed. If that's true, consider updating docs where needed to help make sure we're providing a consistent theory of how this works (to the team now, but for others later).

In general I'm a little worried because we're adding a lot of code here to replace work we might have already finished, adding a bit more with metadata. Is there an approach that wouldn't involve as much replacement, decreasing code diff (and new files, potentially), and remaining in alignment with docs / existing understandings for the project?

d33bs, line 1: How does this file get executed? Consider connecting this back to any existing project execution / pipeline management.

d33bs (approved):

0.download_data/2.download_images.py
MattsonCam: I must have accidentally removed that thinking it was one of my IDE file. I'll add it back in

0.download_data/nbconverted/2.download_image_metadata.py
MattsonCam: Sounds good, the channel names weren't correct before so I'll just mention that

0.download_data/utils/download_images_from_metadata.py
MattsonCam: Ya, I wasn't sure where this should go. It seems more like a utility to me than a necessary step. I'm thinking of putting it in a utils folder and just creating an example of using this utility. This example could then be 4.download_images_from_metadata.py. What do you think?

MattsonCam: > If I'm understanding right, we removed the original image download code and built this new code to do some of that work, in addition to what was needed.

I haven't removed the original image download code yet, but I think it would be straightforward with the util in [4.download_images_from_metadata.py](https://github.com/WayScience/JUMP-single-cell/pull/64/files/7958eea366bfa90457373351e1c536619b2d8f20#diff-9348e59a84cd9b0c43837bb9e1409dc93c74e1008cec288e1b1351bc03efc13b). I created this util so I can parse and download jump image data (and metadata) for the nuclear speckle and anomalyze projects. Do you think it would be worth removing this image download code?

MattsonCam: > In general I'm a little worried because we're adding a lot of code here to replace work we might have already finished, adding a bit more with metadata. Is there an approach that wouldn't involve as much replacement, decreasing code diff (and new files, potentially), and remaining in alignment with docs / existing understandings for the project?

That's true, this code could replace the previous image downloading part, but I don't think that would impact the rest of the pipeline because I could make the image names remain the same. I don't think we need to download the images again though. I initially thought that this code belonged in the JUMP repo even though it is used for two other projects since this code downloads JUMP images using the metadata. Do you think this code should be included in the other repos, or in a different folder in the JUMP repo?

MattsonCam: I think this question can be better answered after the questions in this thread are answered:
https://github.com/WayScience/JUMP-single-cell/pull/64#discussion_r2935620018

0.download_data/nbconverted/3.download_image_metadata.py
MattsonCam: Sounds good

d33bs (approved):
Thanks @MattsonCam, based on our discussion I feel that this LGTM!

0.download_data/2.download_images.py
MattsonCam: After talking i think its better to remove both of them

0.download_data/utils/download_images_from_metadata.py
MattsonCam: we talked about this

MattsonCam: we also talked about this

MattsonCam: we also talked about this

MattsonCam:
Thanks @d33bs , merging now!

PR #65, opened 2026-03-18
Example Image Download using Filtered Metadata

MattsonCam:
Shows an example of how to download images by filtering the image metadata.

### DIFF
--- a/0.download_data/nbconverted/3.download_images.py
+++ b/0.download_data/nbconverted/3.download_images.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# # Download Pilot Images from Filtered Metadata
+# Download JUMP Pilot Images using the filtered manifest metadata.
+
+# In[1]:
+
+
+import runpy
+
+import pandas as pd
+
+
+# ## Inputs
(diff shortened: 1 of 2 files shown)


0.download_data/nbconverted/3.download_images.py
```diff
@@ -0,0 +1,44 @@
...
+import pandas as pd
+
+
+# ## Inputs
+
+# In[2]:
+
+
+mod = runpy.run_path("utils/download_images_from_metadata.py")
+download_images_with_metadata = mod["download_images_with_metadata"]
```
d33bs: Consider importing the function instead of using `runpy`, which is more suited for Python scripts or other non-importable work. While this might work as-is, it's a little unusual so I'd suggest a more "traditional" route.

d33bs (approved):

0.download_data/nbconverted/3.download_images.py
MattsonCam: Sounds good, wanted to try it out, but it makes sense to be consistent

MattsonCam:
Thanks @d33bs , merging now!

